PR #2, opened 2022-10-06
Initial environment set-up and APIs for some transaction and queries

rwalworth:
**Description**:

This PR encompasses the initial environment set-up for the Hedera C++ SDK, as well as some initial work for some transaction and query APIs for interacting with the Hedera network. There is currently no back-end processing for these (i.e. no networking currently to actually submit these transaction/queries to the network), but just provides the main functionality to fill out a transaction/query with the desired fields.

**Related issue(s)**:

**Notes for reviewer**:

Some files/classes were added purely as a means to get the transaction/query objects to build and may not have the entirely correct functionality (see, for example, ContractId.h or FileId.h), but these can be fleshed out as development tasks later; this PR is purely just for pushing up everything that has been initially developed thus far.

These additions have also been untested because the testing framework has not been added yet. This will be added soon.

**Checklist**


rbair23 (commented):
Do we want to check in the protobuf generated content? Or should generating these be part of the build process?

README.md
```diff
@@ -1,2 +1,24 @@
...
+
+## Code of Conduct
+
+This project is governed by the
+[Contributor Covenant Code of Conduct](https://github.com/hashgraph/.github/blob/main/CODE_OF_CONDUCT.md). By
+participating, you are expected to uphold this code of conduct.
+
+## License
+
+[Apache License 2.0](LICENSE)
```
rbair23, line 24: I'd like to clone the repo and try out the build -- should we put the commands in here for how to build and test?

rwalworth: Yeah I can work on adding that today

rwalworth:
> > > Do we want to check in the protobuf generated content? Or should generating these be part of the build process?
> > 
> > 
> > I wasn't sure if it would be worth looking into generating them as part of the build because the process is going to change anyways when the C++ protobufs repo gets up and running. I can definitely look into it though if you think it'll be worthwhile in the short term
> 
> Ya, let's go ahead and omit the generated code from the PR, and just add a step for now in the instructions where the developer has to generate the code before working. I don't like to check in generated code as a general rule

Done, the generated files have been removed and instructions added on how to generate them and add them to the project.

rbair23 (approved):

PR #24, opened 2022-10-13
Remove libsodium in favor of OpenSSL, and make keys into smart pointers

### LINKED ISSUE #23: use OpenSSL instead of libsodium

alittley:
Signed-off-by: Austin Littley <austin.littley@swirldslabs.com>

Resolves #23


sdk/main/include/PublicKey.h
```diff
@@ -20,24 +20,19 @@
...
+  PublicKey(const PublicKey& other) = default;
+  PublicKey& operator=(const PublicKey& other) = default;
+  PublicKey& operator=(const PublicKey&& other) = delete;
 
-  bool operator==(const PublicKey& key) const
-  {
-    (void)key;
-    return true;
-  }
+  virtual proto::Key* toProtobuf() const { return new proto::Key(); }
```
rwalworth, line 35: I know it's just stubbed but would you be able to put this definition in the source file? That way we can remove the inclusion of `<proto/basic_types.ph.b>` from this header file and keep the protobuf files unexposed.

sdk/main/include/PublicKeyFactory.h
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  static std::shared_ptr<PublicKey> fromAliasBytes(const std::string& bytes)
+  {
+    // TODO this implementation is meaningless, only acts as a stub
+    unsigned char keyBytes[20];
+    return std::make_shared<ED25519PublicKey>(ED25519PublicKey(keyBytes));
+  }
+};
```
rwalworth, line 59: Is there a reason you put these factory functions in a separate class as opposed to just in the PublicKey base class?

sdk/main/src/PublicKey.cc
```diff
@@ -1,6 +1,6 @@
 /*-
  *
- * Hedera C++ SDK
+ * Hedera Java SDK
```
rwalworth, line 3:
```suggestion
 * Hedera C++ SDK
```

sdk/main/include/PublicKey.h
alittley: The stub definition is being deleted in my next PR - it becomes pure virtual, since all child classes of PublicKey must implement

```
virtual proto::Key* toProtobuf() const = 0;
```

That said, removing the definition in the header won't let us remove the include - the return type in proto will still need the include, no?

sdk/main/include/PublicKeyFactory.h
alittley: image

It's the difference between these 2 include structures

sdk/main/include/PublicKey.h
rwalworth: You can forward declare the return type, so putting something like `namespace proto { class Key; }` at the top of the file (I've been putting my forward declarations right after the `#include`s) should be all that is needed.

sdk/main/include/PublicKeyFactory.h
rwalworth: Would PublicKey need to depend on the PublicKey child classes if you put the function definitions in a separate source file?

rwalworth: Makes sense, sounds good

rwalworth (approved):

PR #37, opened 2022-10-17
Implement a basic `Transaction` base class

### LINKED ISSUE #11: Implement a basic `Transaction` base class
Implement a basic transaction base class, which will hold information about the transaction's validity, cost, fee, and memo.

Include implementations of the following APIs:
`std::chrono::duration<std::chrono::milliseconds> getTransactionValidDuration() const;`
`T& setTransactionValidDuration(const std::chrono::duration<std::chrono::milliseconds>& validDuration);`
`Hbar getMaxTransactionFee() const;`
`T& setMaxTransactionFee(const Hbar& maxTransactionFee);`
`Hbar getDefaultMaxTransactionFee() const;`
`std::string getTransactionMemo() const;`
`T& setTransactionMemo(const std::string& memo) const;`
`TransactionId getTransactionId() const;`
`T& setTransactionId(const TransactionId& transactionId) const;`

- [x] #17 
- [ ] #20

rwalworth:
This PR adds a `Transaction` base class with base functionality to support submitting cryptography service transactions to a Hedera network. Contains work from #36, so that PR should be approved first.

**Related issue(s)**:

Fixes #11


sdk/main/src/Transaction.cc
```diff
@@ -0,0 +1,65 @@
...
+const std::chrono::duration<double> Transaction<SdkRequestType>::DEFAULT_VALID_TRANSACTION_DURATION =
+  std::chrono::minutes(2);
+
+//-----
+template<typename SdkRequestType>
+const Hbar Transaction<SdkRequestType>::DEFAULT_MAX_TRANSACTION_FEE = Hbar(2LL);
+
+//-----
+template<typename SdkRequestType>
+SdkRequestType& Transaction<SdkRequestType>::setValidTransactionDuration(const std::chrono::duration<double>& duration)
```
alittley, line 35: Don't you risk objects going out of scope by passing const ref into the setters in this file?

rwalworth: The data gets copied into the private member variable and is kept internally, so it wouldn't go out of scope.

alittley (approved):

PR #40, opened 2022-10-18
Implement a basic `AccountCreateTransaction`

### LINKED ISSUE #12: Implement a basic `AccountCreateTransaction`
Implement a basic AccountCreateTransaction type, which will contain all fields needed to create a new account.

Include implementations for the following APIs:
`AccountCreateTransaction& setKey(const Key& key);`
`AccountCreateTransaction& setInitialBalance(const Hbar& initialBalance);`
`AccountCreateTransaction& setReceiverSignatureRequired(bool receiveSignatureRequired);`
`AccountCreateTransaction& setAutoRenewPeriod(const std::chrono::seconds& autoRenewPeriod);`
`AccountCreateTransaction& setMaxAutomaticTokenAssociations(int32_t associations);`
`AccountCreateTransaction& setAccountMemo(const std::string& memo);`
`AccountCreateTransaction& setStakedAccountId(const AccountId& stakedAccountId);`
`AccountCreateTransaction& setStakedNodeId(const int64_t& stakedNodeId);`
`AccountCreateTransaction& setDeclineStakingReward(bool declineStakingReward);`
`AccountCreateTransaction& setAliasKey(const PublicKey& aliasKey);`
`AccountCreateTransaction& setAliasEvmAddress(const EvmAddress& aliasEvmAddress);`
`Key getKey() const;`
`Hbar getInitialBalance() const;`
`bool getReceiverSignatureRequired() const;`
`std::chrono::seconds getAutoRenewPeriod() const;`
`int32_t getMaxAutomaticTokenAssociations() const;`
`std::string getAccountMemo() const;`
`AccountId getStakedAccountId() const;`
`int64_t getStakedNodeId() const;`
`bool getDeclineStakingReward() const;`
`PublicKey getAliasKey() const;`
`EvmAddress getAliasEvmAddress() const;`

- [ ] #16 
- [x] #17
- [x] #18

rwalworth:
This PR cleans up the already-existing `AccountCreateTransaction` and hooks it in with the Transaction base class. Contains work from #37 so that should be approved first.

**Related issue(s)**:

Fixes #12


sdk/main/include/Executable.h
```diff
@@ -0,0 +1,126 @@
...
+#include <chrono>
+#include <vector>
+
+namespace Hedera
+{
+class AccountBalance;
+class AccountBalanceQuery;
+class AccountCreateTransaction;
+class Client;
+class TransactionResponse;
```
alittley, line 34: `TransactionResponse` doesn't appear to be used anywhere. Is it for future use?

sdk/main/include/Executable.h
```diff
@@ -0,0 +1,126 @@
...
+}
+
+namespace proto
+{
+class Query;
+class Response;
+class Transaction;
+class TransactionResponse;
```
alittley, line 42: Also this one

sdk/main/include/Transaction.h
```diff
@@ -20,81 +20,145 @@
...
   /**
-   * Called in freezeWith(Client) just before the transaction body is built. The
-   * intent is for the derived class to assign their data variant to the
-   * transaction body.
+   * Derived from Executable. Construct a transaction protobuf object from this transaction.
+   *
+   * @return The transaction protobuf object that contains this transaction information.
    */
-  virtual void onFreeze(proto::TransactionBody* body) const = 0;
+  virtual proto::Transaction makeRequest() const = 0;
```
alittley, line 133: Do these pure virtual functions need to be declared here? They are already pure virtual in `Executable`, so won't all child classes to this one be forced to implement regardless?

sdk/main/src/AccountBalanceQuery.cc
```diff
@@ -21,113 +21,52 @@
...
-}
+  proto::Query query;
+  proto::CryptoGetAccountBalanceQuery* getAccountBalanceQuery = query.mutable_cryptogetaccountbalance();
 
-//-----
-void
-AccountBalanceQuery::validateChecksums(const Client& client) const
-{
-  if (mAccountId.isValid())
+  if (mAccountId)
```
alittley, line 53: Should the mutual exclusivity of `mAccountId` and `mContractId` be reflected here somehow?

sdk/main/src/AccountCreateTransaction.cc
```diff
@@ -19,233 +19,156 @@
...
-  {
-    mStakedAccountId.getValue().validateChecksum(client);
-  }
-}
-
-//-----
-void AccountCreateTransaction::onFreeze(proto::TransactionBody* body) const
 {
-  body->set_allocated_cryptocreateaccount(build());
+  mMaxTransactionFee = Hbar::from(5LL);
```
alittley, line 36: Is this literal `5` documented anywhere?

sdk/main/src/Executable.cc
```diff
@@ -0,0 +1,83 @@
...
+#include <utility>
+
+namespace Hedera
+{
+//-----
+template<typename SdkRequestType, typename ProtoRequestType, typename ProtoResponseType, typename SdkResponseType>
+SdkResponseType Executable<SdkRequestType, ProtoRequestType, ProtoResponseType, SdkResponseType>::execute(
+  const Client& client)
+{
+  return execute(client, std::chrono::minutes(2L));
```
alittley, line 45: Is this literal documented anywhere?

sdk/main/include/Executable.h
rwalworth: Yup, this shouldn't be here. Removed

sdk/main/include/Executable.h
rwalworth: Removed.

sdk/main/include/Transaction.h
rwalworth: You're right they don't, got copy-paste happy it seems. Removed from here and in `Query` as well.

sdk/main/src/AccountBalanceQuery.cc
rwalworth: I don't think it needs to be, since the set functions prevent both variables being set, unless there's a corner case I'm missing.

sdk/main/src/AccountCreateTransaction.cc
rwalworth: https://docs.hedera.com/guides/mainnet/fees#transaction-and-query-fees

sdk/main/src/Executable.cc
rwalworth: I changed the functionality of this in another PR (using `Client.getRequestTimeout()` instead of the literal) and left a comment in there saying the default is 2 minutes.

alittley (approved):

PR #67, opened 2022-10-21
Use only DER encoding for serialization and deserialization

### LINKED ISSUE #66: Use DER encoding instead of internal encoding to load keys

alittley:
Signed-off-by: Austin Littley <austin.littley@swirldslabs.com>

Resolves #66

### DIFF
--- a/sdk/main/include/ED25519PrivateKey.h
+++ b/sdk/main/include/ED25519PrivateKey.h
@@ -12,10 +12,14 @@ namespace Hedera
 class ED25519PrivateKey : public PrivateKey
 {
 public:
-  ED25519PrivateKey();
-  explicit ED25519PrivateKey(const std::string& privateKeyString);
+  explicit ED25519PrivateKey(EVP_PKEY* keypair);
+
+  static std::shared_ptr<ED25519PrivateKey> generatePrivateKey();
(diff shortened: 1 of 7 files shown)


sdk/main/include/ED25519PrivateKey.h
```diff
@@ -12,11 +12,15 @@ namespace Hedera
 class ED25519PrivateKey : public PrivateKey
 {
 public:
-  ED25519PrivateKey();
-  explicit ED25519PrivateKey(const std::string& privateKeyString);
+  explicit ED25519PrivateKey(EVP_PKEY* keypair);
```
rwalworth, line 15: Would we expect a user to create a `ED25519PrivateKey` from a `EVP_PKEY`?

alittley: No. The reason I had to make the constructor public is to support `std::make_unique` passing in parameters, rather than `std::make_unique` passing in an already constructed object.

Passing an already constructed object makes a copy, rather than directly making a smart pointer. Though now that you mention it, it's probably worth properly implementing a copy constructor to be able to hide the `EVP_PKEY*` constructor. I'll give that a try

rwalworth (approved):

PR #70, opened 2022-10-21
Create CreateAccountExample

### LINKED ISSUE #25: Add example for Create Account
### Problem

Add an example that demonstrates creating an account, gettin the new account account ID, transferring hbars, and getting the account balance. 

Reference example:

- https://github.com/hashgraph/hedera-sdk-java/blob/main/examples/src/main/java/CreateAccountExample.java
- https://github.com/hashgraph/hedera-sdk-java/blob/main/examples/src/main/java/TransferCryptoExample.java

### Solution

Add example

### Alternatives

_No response_

rwalworth:
Add an example that uses the SDK library

**Related issue(s)**:

Fixes #25


sdk/examples/BladeExample.cpp
```diff
@@ -17,27 +17,73 @@
  * limitations under the License.
```
alittley, line 17: Before this merges, I would change the name of the file to describe the example, rather than referencing a particular end user

rwalworth: Yeah I just made this a draft to utilize the GitHub windows build, I'll make it more official when I make this an official PR

PR #110, opened 2022-11-10
00063 authenticated channels

### LINKED ISSUE #63: TLS integration
Figure out what needs to be done with TLS, and do it

alittley:
Resolves #63


sdk/main/include/AccountId.h
```diff
@@ -156,9 +156,21 @@ class AccountId
   /**
    * The account ID number.
    */
-  std::optional<uint64_t> mAccountNum;
+  uint64_t mAccountNum;
 };
 
 } // namespace Hedera
 
+namespace std
```
rwalworth, line 164: There a reason this is in the `std` namespace instead of the `Hedera`?

sdk/main/include/IPv4Address.h
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * @return string representing the address
+   */
+  [[nodiscard]] std::string toString() const;
+
+private:
+  /**
+   * The first octet of the address
+   */
+  unsigned int mOctet1;
```
rwalworth, line 65: Should these be `unsigned char`s since I'm assuming they can only be a byte?

sdk/main/include/TLSBehavior.h
```diff
@@ -0,0 +1,44 @@
...
+#ifndef HEDERA_SDK_CPP_TLSBEHAVIOR_H
+#define HEDERA_SDK_CPP_TLSBEHAVIOR_H
+
+namespace Hedera
+{
+
+/**
+ * Enum representing different modes of network operation
+ */
+enum TLSBehavior
```
rwalworth, line 30: Are these going to be the only two possible options for `TLSBehavior`? If so, would it make sense just to use a boolean instead?

sdk/main/include/NodeAddress.h
```diff
@@ -21,27 +21,102 @@
...
+/**
+ * Class containing all information related the the address(es) of a node
+ */
 class NodeAddress
 {
 public:
+  inline static int PORT_MIRROR_PLAIN = 5600;
+  inline static int PORT_MIRROR_TLS = 443;
+  inline static int PORT_NODE_PLAIN = 50211;
+  inline static int PORT_NODE_TLS = 50212;
```
rwalworth, line 46: These can probably be `const`.

sdk/main/include/TransferTransaction.h
```diff
@@ -66,7 +66,7 @@ class TransferTransaction : public Transaction<TransferTransaction>
    * @param amount    The amount to transfer.
    * @return Reference to this TransferTransaction object.
    */
-  TransferTransaction& addUnapprovedHbarTransfer(const AccountId& accountId, const Hbar& amount);
+  TransferTransaction& addUnapprovedHbarTransfer(std::shared_ptr<AccountId> accountId, const Hbar& amount);
```
rwalworth, line 69:
```suggestion
  TransferTransaction& addUnapprovedHbarTransfer(const std::shared_ptr<AccountId>& accountId, const Hbar& amount);
```

sdk/main/src/IPv4Address.cc
```diff
@@ -0,0 +1,110 @@
...
+    };
+
+    return outputAddress;
+  }
+
+  // if input string isn't bytes, it must be made up of number characters, with 4 octets separated by the '.' character
+  if (std::count(inputString.begin(), inputString.end(), '.') != 3)
+  {
+    // todo THROW
+    return { 0, 0, 0, 0 };
```
rwalworth, line 58:
```suggestion
    throw std::invalid_argument("Input IPv4Address is malformed");
```

Can probably use this for all the other `// todo THROW`s you have in this function.

sdk/main/src/Node.cc
```diff
@@ -51,4 +54,50 @@ void Node::shutdown()
...
 }
 
+bool Node::checkChannelInitialized()
+{
+  if (mChannel.getInitialized() || tryInitializeChannel())
+  {
+    return true;
+  }
+
+  return false;
```
rwalworth, line 64:
```suggestion
  return mChannel.getInitialized() || tryInitializeChannel();
```

sdk/main/include/Channel.h
```diff
@@ -67,8 +60,19 @@ class Channel
    *
    * @param url The URL and port of the gRPC service with which this channel
    *            should communicate.
+   * @param requireTLS initializes the channel with TLS if true, otherwise initializes an unsecured channel
+   * @param nodeCertHash the sha-384 hash of the node certificate chain. only relevant if requireTLS is true
+   *
+   * @return true if channel was successfully initialized, otherwise false
    */
-  void initChannel(const std::string& url);
+  bool initializeChannel(const std::string& url, bool requireTLS, const std::string& nodeCertHash = "");
```
rwalworth, line 68: Could this function just check if `nodeCertHash` is empty? Or is `requireTLS` required for something else that I'm missing?

sdk/tests/NodeAddressBookTest.cc
```diff
@@ -0,0 +1,40 @@
...
+
+using namespace Hedera;
+
+class NodeAddressBookTest : public ::testing::Test
+{
+protected:
+  void SetUp() override {}
+};
+
+TEST_F(NodeAddressBookTest, readAddressBookResource)
```
rwalworth, line 32: What is this test actually testing? Should it check each address and make sure they're parsed correctly?

sdk/main/include/TLSBehavior.h
alittley: More options are expected in the future, similar to https://vertx.io/docs/apidocs/io/vertx/pgclient/SslMode.html

sdk/main/include/AccountId.h
alittley: It's my understanding that it's required to be this way to override the standard hash function

sdk/tests/NodeAddressBookTest.cc
alittley: Oops, this was just my debugging entry point. I'll remove the file for now, since I haven't written any meaningful tests

sdk/main/include/Channel.h
alittley: I made separate functions to get rid of the yuckiness

rwalworth (approved):

PR #120, opened 2022-11-18
Create unit tests for `AccountId`

### LINKED ISSUE #77: Create unit tests for `AccountId`

rwalworth:
This PR adds unit tests for `AccountId`.

**Related issue(s)**:

Fixes #77 

**Notes for reviewer**:
This contains work from #119, so that PR should be reviewed and approved first.

**Checklist**


sdk/main/src/AccountId.cc
```diff
@@ -41,9 +41,10 @@ AccountId::AccountId(const uint64_t& shard, const uint64_t& realm, const uint64_
 //-----
 AccountId::AccountId(const std::string& str)
 {
+  size_t numDots = 0ULL;
   for (char c : str)
   {
-    if (!isdigit(c) && c != '.')
+    if ((!isdigit(c) && c != '.') || (c == '.' && ++numDots > 2))
```
alittley, line 47: I would suggest removing this `numDots` check in the loop, and just asserting below the loop that `numDots == 2`

As it currently stands, the following test assertion *should* throw an error, but doesn't:
```
EXPECT_THROW(AccountId(testShardNumStr + testRealmNumStr + '.' + testAccountNumStr), std::invalid_argument);
```

sdk/tests/AccountIdTest.cc
```diff
@@ -0,0 +1,128 @@
...
+  EXPECT_EQ(protoAccountId->shardnum(), getTestShardNum());
+  EXPECT_EQ(protoAccountId->realmnum(), getTestRealmNum());
+  EXPECT_EQ(protoAccountId->accountnum(), getTestAccountNum());
+
+  const uint64_t adjustment = 3ULL;
+  const uint64_t newShard = getTestShardNum() + adjustment;
+  const uint64_t newRealm = getTestRealmNum() - adjustment;
+  const uint64_t newContract = getTestAccountNum() * adjustment;
+
+  protoAccountId->set_shardnum(static_cast<int64_t>(newShard));
```
alittley, line 120: Seeing this test case brings to mind - 

why are `AccountId` member variables `uint_64`, if the protobuf fields are `int64_t`? As it stands, there are a bunch of `uint_64` values which could be successfully set on the class, that would produce undefined behavior with this static cast

sdk/main/src/AccountId.cc
rwalworth: It would throw an error because it checks to make sure that the `shardStr`, `realmStr`, and `accountStr` aren't empty, which `realmStr` would be since the substring between `firstDot` and `secondDot` (since they're the same dot) would be size=0. This is a lot shoved into one line of code though, I can spread it out a bit.

sdk/tests/AccountIdTest.cc
```diff
@@ -0,0 +1,128 @@
...
+  [[nodiscard]] inline const uint64_t& getTestRealmNum() const { return mRealmNum; }
+  [[nodiscard]] inline const uint64_t& getTestAccountNum() const { return mAccountNum; }
+
+private:
+  const uint64_t mShardNum = 8ULL;
+  const uint64_t mRealmNum = 9ULL;
+  const uint64_t mAccountNum = 10ULL;
+};
+
+TEST_F(AccountIdTest, DefaultConstructaccountId)
```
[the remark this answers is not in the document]
rwalworth: I have a feeling this got overwritten when I was maybe doing a replace all or something, I prefer the `DefaultConstructAccountId` as well. I'll change it.

sdk/main/src/AccountId.cc
alittley: I tried the test and it didn't fail - nothing is thrown by that input

sdk/tests/AccountIdTest.cc
rwalworth: For some reason the protobuf fields aren't generated as `uint64_t` even though they should be nonnegative. I'm not sure if this is a limitation of protobuf or what, but I felt it should be enforced in the SDK that these numbers can't be negative. Maybe I can just put a comment for the class to saying that 2^64 - 1 is the max value for each of these fields?

sdk/main/src/AccountId.cc
rwalworth: Sorry yeah I was saying the constructor would throw, therefore not causing the test to fail.

alittley: But it doesn't throw? Unless I'm misunderstanding what you're saying

Calling `AccountId("66.77")` succeeds, with a resulting shard, realm, account number of `66, 77, 77`

sdk/tests/AccountIdTest.cc
alittley: If the protobuf has `int64`, I would say the underlying object should as well. Non-negative values could be enforced by `if`s in the constructors and setters.

IMO a comment isn't enough, since they can be ignored. If the underlying class values are kept as `uint64`, there would need to be an `if` block anyway, to make sure the value isn't too big. Either way, a manual check is necessary

sdk/main/src/AccountId.cc
rwalworth: Oh no I totally misunderstood what you were saying, my bad. Yeah you are correct, that case is not caught. I'll update the test with that and directly check for two dots.

sdk/tests/AccountIdTest.cc
rwalworth: I'll add some throws for input numbers that are too big.

I personally like keeping the `uint64_t` type just to enforce in the code that the type should be nonnegative. The way I see it, either way `if` blocks will have to be added to verify values are within range (either `int64_t` is nonnegative or `uint64_t` isn't too big) and comments will be added to specify ranges to the user. The `uint64_t` type I feel better exemplifies how the value should be seen and used. With that said my opinions not super strong so I'm willing to change it if it causes heartburn.

alittley: I'm fine with this way :+1:

A couple things remaining-
1. the string constructor also needs the max value checks
2. I would personally add some tests to make sure the checks throughout `AccountId` are hit, passing in `int64 max` and `int64 max +1` to verify. Up to you, though

rwalworth: Tests added

sdk/main/src/AccountId.cc
```diff
@@ -60,42 +60,91 @@ AccountId::AccountId(const uint64_t& shard, const uint64_t& realm, const uint64_
...
-  if (numDots != 2ULL)
+  // Make sure all numbers where constructed
+  if (numbers.at(0).empty() || numbers.at(1).empty() || numbers.at(2).empty())
   {
     throw std::invalid_argument("AccountId string is malformed");
   }
 
-  const size_t firstDot = str.find_first_of('.');
-  const size_t secondDot = str.find_last_of('.');
+  // Translate out_or_range exception to invalid_argument to allow for more descriptive exception
```
alittley: "out_*of*_range" exception

sdk/tests/AccountIdTest.cc
```diff
@@ -91,6 +102,7 @@ TEST_F(AccountIdTest, ConstructFromString)
   EXPECT_THROW(AccountId("abc"), std::invalid_argument);
   EXPECT_THROW(AccountId("o.o.e"), std::invalid_argument);
   EXPECT_THROW(AccountId("0.0.1!"), std::invalid_argument);
+  EXPECT_THROW(AccountId(testNumTooBigStr + '.' + testNumTooBigStr + '.' + testNumTooBigStr), std::invalid_argument);
```
alittley, line 106: Since the error checking is split into different blocks in the original code, I would add an `EXPECT_THROW` for each individual string being too big. Currently, you'll hit the same error block each time and early out

sdk/tests/AccountIdTest.cc
```diff
@@ -59,13 +65,18 @@ TEST_F(AccountIdTest, ConstructWithShardRealmAccountNum)
   EXPECT_EQ(accountId.getShardNum(), getTestShardNum());
   EXPECT_EQ(accountId.getRealmNum(), getTestRealmNum());
   EXPECT_EQ(accountId.getAccountNum(), getTestAccountNum());
+
+  EXPECT_THROW(AccountId(getTestNumTooBig(), getTestRealmNum(), getTestAccountNum()), std::invalid_argument);
+  EXPECT_THROW(AccountId(getTestShardNum(), getTestNumTooBig(), getTestAccountNum()), std::invalid_argument);
+  EXPECT_THROW(AccountId(getTestShardNum(), getTestRealmNum(), getTestNumTooBig()), std::invalid_argument);
 }
 
 TEST_F(AccountIdTest, ConstructFromString)
```
alittley, line 75: Currently, all of your successful fromString tests have only a single digit character. Since this deals with character parsing, I think it makes sense to also test with multicharacter numbers

sdk/main/src/AccountId.cc
rwalworth: Fixed.

sdk/tests/AccountIdTest.cc
rwalworth: Fixed.

sdk/tests/AccountIdTest.cc
rwalworth: I changed the test realm and account numbers to be multi-character numbers.

rwalworth:
> IMO a better way to fix the most vexing parse is by using curly braces instead of parenthesis for the constructors

Yup, agreed. Updated.

PR #133, opened 2022-12-08
Fix SDK build under Windows OS environment

### LINKED ISSUE #131: Fix SDK build under Windows OS environment
### Problem

The Windows build is currently broken and needs to be fixed to support the end-user requirements.

### Solution

- Fix the build
- Test using the example program

### Alternatives

None

nathanklick:
## Description

Resolves the Windows OS compilation and linker issues.

### Related Issues 

- Closes #131


rwalworth (approved):

alittley (approved):
Changes look good, and build was successful

david-bakin-sl:
Had a couple of problems building on my system using VS2022 (that is, using the VS2022 generator).

1. Couldn't build the Debug configuration (`cmake --build build --config Debug`) (which is the default config).  Reason is that `CMakeLists.txt` specifies linking against static libraries, which includes `MSVCRT` (i.e., the C++ runtimes), but the HAPI library pulled down by `HederaAPI.cmake` is built against Release libraries, thus a conflict.  The "proper" fix is probably to supply downloads of HAPI _both_ for Release and Debug, other alternatives would be building HAPI here, or trying to specify `/NODEFAULTLIB:library` as suggested by the linker (though they may not actually work for here, we'd have to check).  _Building_ the Release configuration works (but see next ...)  I didn't try ninja because the linker will run in to the same problem.
2.  `ctest` failed.  (First, I had to use the command `ctest --C Release` - the `-C` option is missing from the `README` but is required, that should be fixed in the `README`.) Log attached.  I'm not familiar with this stuff so someone else needs to look at this and advise.

[sdk-ctest-fail.log](https://github.com/hashgraph/hedera-sdk-cpp/files/10211930/sdk-ctest-fail.log)

Issue #152, opened 2022-12-20
Remove remaining prints on failure


alittley:
- There are some instances of `cout` messages indicating failures
- These should be replaced by exceptions

deyanzz:
I found only one usage of `cout` message as a failure.
My changes are introduced in [PR 60](https://github.com/hashgraph/hedera-sdk-cpp/pull/160).

deyanzz:
[PR 160](https://github.com/hashgraph/hedera-sdk-cpp/pull/160) was approved and merged into `main` branch.

PR #160, opened 2023-01-05
Remove remaining prints on failure and updated tests in ED25519PublicKeyTest

deyanzz:
**Description**:
I found only one instance of `cout` message, which indicates a failure.
According to issue [#152](https://github.com/hashgraph/hedera-sdk-cpp/issues/152) it has be replaced by an exception.

**Related issue(s)**: [Issue #152](https://github.com/hashgraph/hedera-sdk-cpp/issues/152)

Fixes #
The fix is only in `ED25519PublicKey.cc`, where method `ED25519PublicKey::verifySignature` was modified.
From now on, when calling `ED25519PublicKey::verifySignature` and `verificationResult <= 0`, then it will throw an `std::runtime_error` with the error message.
Updated the following tests in `ED25519PublicKeyTest.cc`:
 - `VerifySignatureAgainstModifiedBytes`
 - `VerifyArbitrarySignature`
 - `VerifyEmptySignature`
 - `VerifyEmptyMessage`

**Notes for reviewer**:
With the latest changes, method `ED25519PublicKey::verifySignature` is not possible to return **FALSE** as a result. Maybe this is not correct at all, so need to be discussed.

**Checklist**


sdk/main/src/ED25519PublicKey.cc
```diff
@@ -112,7 +112,8 @@ bool ED25519PublicKey::verifySignature(const std::vector<unsigned char>& signatu
 
   if (verificationResult <= 0)
   {
-    std::cout << "Failed to verify signature with code [" << verificationResult << ']' << std::endl;
+    std::string message = "Failed to verify signature with code [" + std::to_string(verificationResult) + ']';
+    throw std::runtime_error(message.c_str());
```
rwalworth, line 116: You need to free the `messageDigestContext` here before throwing. Something like `EVP_MD_CTX_free(messageDigestContext)`.

sdk/tests/ED25519PublicKeyTest.cc
```diff
@@ -107,39 +107,39 @@ TEST_F(ED25519PublicKeyTest, VerifySignatureAgainstModifiedBytes)
   const std::vector<unsigned char> signature = getTestPrivateKey()->sign({ 0x1, 0x2, 0x3 });
   const std::vector<unsigned char> modifiedBytes = { 0x1, 0x2, 0x3, 0x4 };
 
-  EXPECT_FALSE(getTestPublicKeyFromPrivate()->verifySignature(signature, modifiedBytes));
-  EXPECT_FALSE(getTestPublicKeyFromString()->verifySignature(signature, modifiedBytes));
-  EXPECT_FALSE(getTestPublicKeyFromProtobuf()->verifySignature(signature, modifiedBytes));
+  EXPECT_ANY_THROW(getTestPublicKeyFromPrivate()->verifySignature(signature, modifiedBytes));
+  EXPECT_ANY_THROW(getTestPublicKeyFromString()->verifySignature(signature, modifiedBytes));
+  EXPECT_ANY_THROW(getTestPublicKeyFromProtobuf()->verifySignature(signature, modifiedBytes));
```
rwalworth, line 112: Since only a `std::runtime_error` is expected to be thrown here, you could do `EXPECT_THROW(getTestPublicKeyFrom<variant>()->verifySignature(signature, modifiedBytes), std::runtime_error);` or something to that effect instead. This goes for all updated test assertions in this PR.

rwalworth (changes requested):

sdk/main/src/ED25519PublicKey.cc
deyanzz: Thanks for noticing it. I fixed it in 04f7036023e74d9f62f24cd49f49801eabab2d2e.

sdk/tests/ED25519PublicKeyTest.cc
deyanzz: All of the tests have been updated:
 - test `VerifySignatureAgainstModifiedBytes` - in 370451f0a120d5bf95b69150da2e7c5dab62a529;
 - test `VerifyArbitrarySignature` - in 712931257071667967a4a592fb31e0e6394517a2;
 - test `VerifyEmptySignature` - in 1fe2f718d534b9ed622c2b41a316b320e5bb7db7;
 - test `VerifyEmptyMessage` - in d708b8d5ce69ee84751ea3cd2a4d52f5f2dd4b21;

rwalworth (approved):

PR #162, opened 2023-01-11
Build OpenSSL from source

### LINKED ISSUE #143: Build OpenSSL from source
- `keccak256` support has been merged into `master`, but won't be available until OpenSSL v3.2 at the earliest
- building from master is worth avoiding relying on a separate library for `keccak256`

rwalworth:
**Description**:
This PR adds the `master` OpenSSL repo to the SDK so that needed features that are not in a formal OpenSSL release (i.e. `keccak256`) can be utilized by the SDK. This has been confirmed as working locally and changes to the SDK that required `master` OpenSSL will be in separate PRs.

**Related issue(s)**:

Fixes #143 

**Notes for reviewer**:
The first build with this change will be longer than usual, as the SDK has to build and install `master` OpenSSL.

**Checklist**


BuildOpenSSL.cmake
```diff
@@ -0,0 +1,50 @@
+include(ExternalProject)
+set(OPENSSL_SOURCE_DIR ${CMAKE_CURRENT_BINARY_DIR}/_deps/openssl-src) # default path by CMake
+set(OPENSSL_INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/openssl)
+set(OPENSSL_INCLUDE_DIR ${OPENSSL_INSTALL_DIR}/include)
+
+if (WIN32)
+    set(OPENSSL_CONFIGURE_COMMAND perl ${OPENSSL_SOURCE_DIR}/Configure VC-WIN64A)
+    set(OPENSSL_BUILD_COMMAND nmake /S)
+    set(OPENSSL_INSTALL_COMMAND nmake install /S)
+    set(OPENSSL_LIBRARY_EXTENSION .lib)
```
rwalworth, line 10: @nathanklick these changes here involve the installed `perl` and `nmake` on our windows build machines. I had to update the %PATH% variable in order to access these executables. I'm not sure if this is the right way to do it so any feedback here would be appreciated.

alittley:
@rwalworth Have you tested that this builds and tests pass on Windows?

rwalworth:
> @rwalworth Have you tested that this builds and tests pass on Windows? (I see you have commits about that, so I'm assuming yes)

Yeah I built and tested on my local windows machine and it all looks good. Only caveat with building on windows however (and I left a comment concerning this) is that `perl` and `nmake` need to be installed and cmake needs to properly reference them (I added the needed executables to my %PATH% variable, which I'm guessing is probably the incorrect way to do it), so I'm hoping for a little guidance from @nathanklick on what to do there.

nathanklick (approved):

PR #169, opened 2023-01-18
Swap sha256 hashing for keccak256

### LINKED ISSUE #144: Swap `sha256` hashing for `keccak256`
- ECDSA signatures currently are made on the `sha256` of the input message
- They must instead be made on the `keccak256` hash, as in ethereum
    - NOTE: `keccak256` != `sha3-256`. It's a pre standard version of the hashing function (see [this OpenSSL issue](https://github.com/openssl/openssl/issues/13033) for more detail)
- This can only be done once #143 is complete

rwalworth:
**Description**:
This PR incorporates the new functionality granted from #143 and correctly implements `ECDSA` signing and verification.

**Related issue(s)**:

Fixes #144 

**Checklist**


sdk/main/src/ECDSAPublicKey.cc
```diff
@@ -99,22 +99,79 @@ std::unique_ptr<PublicKey> ECDSAPublicKey::clone() const
...
+
   EVP_MD_CTX* messageDigestContext = EVP_MD_CTX_new();
 
   if (!messageDigestContext)
   {
     throw std::runtime_error(internal::OpenSSLHasher::getOpenSSLErrorMessage("EVP_MD_CTX_new"));
   }
 
-  if (EVP_DigestVerifyInit(messageDigestContext, nullptr, EVP_sha256(), nullptr, mPublicKey) <= 0)
+  OSSL_LIB_CTX* libraryContext = OSSL_LIB_CTX_new();
```
alittley, line 157: I would check `libraryContext` and `messageDigest` for nullity right after creating them (I believe we do this everywhere else, too). I'm not sure if subsequent calls would just fail if they are null, or if you would get a segfault.

sdk/main/src/ECDSAPrivateKey.cc
```diff
@@ -149,7 +149,14 @@ std::vector<unsigned char> ECDSAPrivateKey::sign(const std::vector<unsigned char
     throw std::runtime_error(internal::OpenSSLHasher::getOpenSSLErrorMessage("EVP_MD_CTX_new"));
   }
 
-  if (EVP_DigestSignInit(messageDigestContext, nullptr, EVP_sha256(), nullptr, mKeypair) <= 0)
+  OSSL_LIB_CTX* libraryContext = OSSL_LIB_CTX_new();
```
alittley, line 152: Any changes you decide the make to `PublicKey` should happen here too

alittley (approved):

sdk/main/src/ECDSAPublicKey.cc
rwalworth: Added

sdk/main/src/ECDSAPrivateKey.cc
rwalworth: Added

PR #191, opened 2023-02-02
Use `std::unique_ptr` to manage OpenSSL memory allocations when able

### LINKED ISSUE #173: Replace raw OpenSSL pointer handling with `std::unique_ptr` implementation with custom deleter
### Problem

It's generally not great C++ practice to directly handle resources and RAII should be preferred. Functions involving OpenSSL handle raw pointers, but there may be a way to implement that same functionality using `std::unique_ptr` and a custom deleter.

### Solution

Something like: `std::unique_ptr<EVP_PKEY, void (*)(EVP_PKEY)> ptr(EVP_PKEY_new(), &EVP_PKEY_free);`. This will allow the `std::unique_ptr` to properly handle the deletion of the OpenSSL elements. Each OpenSSL object generally has a `new` and a `free` function of some sort, but if the `free` function is fed into a `std::unique_ptr` for the element that is `new`'d, it should be able to handle deletion how a `std::unique_ptr` normally does.

### Alternatives

_No response_

rwalworth:
**Description**:
This PR consolidates OpenSSL memory management by utilizing `std::unique_ptr` with a custom deleter to automatically handle freeing memory when exiting scope.

**Related issue(s)**:

Fixes #173 

**Checklist**

### DIFF
--- a/sdk/main/include/ECDSAsecp256k1PrivateKey.h
+++ b/sdk/main/include/ECDSAsecp256k1PrivateKey.h
@@ -23,6 +23,7 @@
 #include "ECDSAsecp256k1PublicKey.h"
 #include "MnemonicBIP39.h"
 #include "PrivateKey.h"
+#include "impl/OpenSSLObjectWrapper.h"
 
 #include <memory>
 #include <openssl/crypto.h>
@@ -144,29 +145,27 @@ public:
 
 private:
   /**
-   * Create an OpenSSL key object from a byte vector representing an ECDSAsecp256k1PrivateKey.
+   * Create a wrapped OpenSSL key object from a byte vector representing an ECDSAsecp256k1PrivateKey.
    *
    * @param keyBytes The bytes representing a ECDSAsecp256k1PrivateKey.
-   * @return A pointer to a newly created OpenSSL keypair object.
+   * @return The newly created wrapped OpenSSL keypair object.
    */
-  static std::unique_ptr<EVP_PKEY, void (*)(EVP_PKEY*)> bytesToPKEY(const std::vector<unsigned char>& keyBytes);
+  static internal::OpenSSL_EVP_PKEY bytesToPKEY(const std::vector<unsigned char>& keyBytes);
 
   /**
-   * Construct from an OpenSSL keypair object.
+   * Construct from a wrapped OpenSSL keypair object.
    *
-   * @param keypair A pointer to the underlying OpenSSL keypair object from which to construct this
-   *                ECDSAsecp256k1PrivateKey.
+   * @param keypair The wrapped OpenSSL keypair object from which to construct this ECDSAsecp256k1PrivateKey.
    */
-  explicit ECDSAsecp256k1PrivateKey(std::unique_ptr<EVP_PKEY, void (*)(EVP_PKEY*)>&& keypair);
+  explicit ECDSAsecp256k1PrivateKey(internal::OpenSSL_EVP_PKEY&& keypair);
 
   /**
-   * Construct from an OpenSSL keypair object and a chaincode.
+   * Construct from a wrapped OpenSSL keypair object and a chaincode.
    *
-   * @param keypair   A pointer to the underlying OpenSSL keypair.
+   * @param keypair   The wrapped OpenSSL keypair.
    * @param chainCode The new ECDSAsecp256k1PrivateKey's chain code.
    */
-  ECDSAsecp256k1PrivateKey(std::unique_ptr<EVP_PKEY, void (*)(EVP_PKEY*)>&& keypair,
-                           std::vector<unsigned char> chainCode);
+  ECDSAsecp256k1PrivateKey(internal::OpenSSL_EVP_PKEY&& keypair, std::vector<unsigned char> chainCode);
 
   /**
    * Get the byte representation of the ECDSAsecp256k1PublicKey that corresponds to this ECDSAsecp256k1PrivateKey.
@@ -176,9 +175,9 @@ private:
   [[nodiscard]] std::vector<unsigned char> getPublicKeyBytes() const;
 
   /**
-   * A pointer to the underlying OpenSSL keypair.
+   * The wrapped OpenSSL keypair.
    */
-  std::unique_ptr<EVP_PKEY, void (*)(EVP_PKEY*)> mKeypair = { nullptr, &EVP_PKEY_free };
+  internal::OpenSSL_EVP_PKEY mKeypair;
 
   /**
    * A pointer to the ECDSAsecp256k1PublicKey object that corresponds to this ECDSAsecp256k1PrivateKey.
--- a/sdk/main/include/ECDSAsecp256k1PrivateKey.h
+++ b/sdk/main/include/ECDSAsecp256k1PrivateKey.h
@@ -26,6 +26,7 @@
 
 #include <memory>
 #include <openssl/crypto.h>
+#include <openssl/evp.h>
 #include <string>
 #include <vector>
 
@@ -44,13 +45,9 @@ public:
   ECDSAsecp256k1PrivateKey() = delete;
 
   /**
-   * Destructor
-   */
-  ~ECDSAsecp256k1PrivateKey() override;
-
-  /**
    * Copy constructor and copy assignment operator can throw std::runtime_error if OpenSSL serialization fails.
    */
+  ~ECDSAsecp256k1PrivateKey() override = default;
   ECDSAsecp256k1PrivateKey(const ECDSAsecp256k1PrivateKey& other);
   ECDSAsecp256k1PrivateKey& operator=(const ECDSAsecp256k1PrivateKey& other);
   ECDSAsecp256k1PrivateKey(ECDSAsecp256k1PrivateKey&& other) noexcept;
@@ -147,12 +144,12 @@ public:
 
 private:
   /**
-   * Create an OpenSSL keypair object from a byte vector representing an ECDSAsecp256k1PrivateKey.
+   * Create an OpenSSL key object from a byte vector representing an ECDSAsecp256k1PrivateKey.
    *
    * @param keyBytes The bytes representing a ECDSAsecp256k1PrivateKey.
    * @return A pointer to a newly created OpenSSL keypair object.
    */
-  static EVP_PKEY* bytesToPKEY(const std::vector<unsigned char>& keyBytes);
+  static std::unique_ptr<EVP_PKEY, void (*)(EVP_PKEY*)> bytesToPKEY(const std::vector<unsigned char>& keyBytes);
 
   /**
    * Construct from an OpenSSL keypair object.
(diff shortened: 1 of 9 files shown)


deyanzz (approved):
LGTM. Tested locally and all of the tests passed.

sdk/main/include/ECDSAsecp256k1PrivateKey.h
```diff
@@ -147,28 +144,29 @@ class ECDSAsecp256k1PrivateKey : public PrivateKey
...
 private:
   /**
-   * Create an OpenSSL keypair object from a byte vector representing an ECDSAsecp256k1PrivateKey.
+   * Create an OpenSSL key object from a byte vector representing an ECDSAsecp256k1PrivateKey.
    *
    * @param keyBytes The bytes representing a ECDSAsecp256k1PrivateKey.
    * @return A pointer to a newly created OpenSSL keypair object.
    */
-  static EVP_PKEY* bytesToPKEY(const std::vector<unsigned char>& keyBytes);
+  static std::unique_ptr<EVP_PKEY, void (*)(EVP_PKEY*)> bytesToPKEY(const std::vector<unsigned char>& keyBytes);
```
alittley, line 152: Have you considered using `decltype` for the destructor, rather than explicitly typing the type out?

sdk/main/include/ECDSAsecp256k1PrivateKey.h
```diff
@@ -147,28 +144,29 @@ class ECDSAsecp256k1PrivateKey : public PrivateKey
...
```
alittley, line 153: Have you considered making a wrapper class around EVP_PKEY (and any other openssl object you follow this custom deleter pattern for)? The custom deleter logic could be relegated to the wrapper classes and thoroughly explained in just 1 spot

sdk/main/include/ECDSAsecp256k1PrivateKey.h
rwalworth: No I didn't consider that, can add it though. I think I'll add an alias for these classes as well to clean up a bit.

sdk/main/include/ECDSAsecp256k1PrivateKey.h
rwalworth: Is `std::unique_ptr` not the wrapper class? I don't know if I fully understand what you're suggesting

alittley: That was the thought - a c++ class containing only the unique pointer, which knows how to safely destruct. May or may not be a good idea

rwalworth: Pushed up an implementation, let me know what you think

sdk/main/include/impl/OpenSSLObjectWrapper.h
```diff
@@ -0,0 +1,250 @@
...
+/**
+ * Wrapper class for the OpenSSL ECDSA_SIG object.
+ */
+class OpenSSL_ECDSA_SIG : public OpenSSLObjectWrapper<ECDSA_SIG, decltype(&ECDSA_SIG_free)>
+{
+public:
+  /**
+   * Default constructor sets the ECDSA_SIG_free deleter function.
+   */
+  OpenSSL_ECDSA_SIG()
```
alittley, line 82: Any reason we need a default constructor, rather than just deleting it?

sdk/main/include/impl/OpenSSLObjectWrapper.h
```diff
@@ -0,0 +1,250 @@
...
+    : OpenSSLObjectWrapper(nullptr, &ECDSA_SIG_free)
+  {
+  }
+
+  /**
+   * Construct with the input ECDSA_SIG and its ECDSA_SIG_free deleter function.
+   *
+   * @param evpPkey The ECDSA_SIG OpenSSL object to wrap.
```
alittley, line 90: wrong arg name

sdk/main/include/impl/OpenSSLObjectWrapper.h
```diff
@@ -0,0 +1,250 @@
...
+   */
+  OpenSSL_EVP_MD()
+    : OpenSSLObjectWrapper(nullptr, &EVP_MD_free)
+  {
+  }
+
+  /**
+   * Construct with the input EVP_MD and its EVP_MD_free deleter function.
+   *
+   * @param evpPkey The EVP_MD OpenSSL object to wrap.
```
alittley, line 115: wrong arg name

sdk/main/include/impl/OpenSSLObjectWrapper.h
```diff
@@ -0,0 +1,250 @@
...
+   */
+  OpenSSL_EVP_MD_CTX()
+    : OpenSSLObjectWrapper(nullptr, &EVP_MD_CTX_free)
+  {
+  }
+
+  /**
+   * Construct with the input EVP_MD_CTX and its EVP_MD_CTX_free deleter function.
+   *
+   * @param evpPkey The EVP_MD_CTX OpenSSL object to wrap.
```
alittley, line 140: arg name

sdk/main/include/impl/OpenSSLObjectWrapper.h
```diff
@@ -0,0 +1,250 @@
...
+   */
+  OpenSSL_EVP_PKEY()
+    : OpenSSLObjectWrapper(nullptr, &EVP_PKEY_free)
+  {
+  }
+
+  /**
+   * Construct with the input EVP_PKEY and its EVP_PKEY_free deleter function.
+   *
+   * @param evpPkey The EVP_PKEY OpenSSL object to wrap.
```
alittley, line 165: arg name

sdk/main/include/impl/OpenSSLObjectWrapper.h
```diff
@@ -0,0 +1,250 @@
...
+   */
+  OpenSSL_EVP_PKEY_CTX()
+    : OpenSSLObjectWrapper(nullptr, &EVP_PKEY_CTX_free)
+  {
+  }
+
+  /**
+   * Construct with the input EVP_PKEY_CTX and its EVP_PKEY_CTX_free deleter function.
+   *
+   * @param evpPkey The EVP_PKEY_CTX OpenSSL object to wrap.
```
alittley, line 190: arg name

sdk/main/include/impl/OpenSSLObjectWrapper.h
```diff
@@ -0,0 +1,250 @@
...
+   */
+  OpenSSL_OSSL_LIB_CTX()
+    : OpenSSLObjectWrapper(nullptr, &OSSL_LIB_CTX_free)
+  {
+  }
+
+  /**
+   * Construct with the input OSSL_LIB_CTX and its OSSL_LIB_CTX_free deleter function.
+   *
+   * @param evpPkey The OSSL_LIB_CTX OpenSSL object to wrap.
```
alittley, line 215: arg name

sdk/main/include/impl/OpenSSLObjectWrapper.h
```diff
@@ -0,0 +1,250 @@
...
+   */
+  OpenSSL_OSSL_DECODER_CTX()
+    : OpenSSLObjectWrapper(nullptr, &OSSL_DECODER_CTX_free)
+  {
+  }
+
+  /**
+   * Construct with the input OSSL_DECODER_CTX and its OSSL_DECODER_CTX_free deleter function.
+   *
+   * @param evpPkey The OSSL_DECODER_CTX OpenSSL object to wrap.
```
alittley, line 240: arg name

sdk/main/include/ECDSAsecp256k1PrivateKey.h
```diff
@@ -43,14 +45,10 @@ class ECDSAsecp256k1PrivateKey : public PrivateKey
...
 
-  /**
-   * Destructor
-   */
-  ~ECDSAsecp256k1PrivateKey() override;
-
   /**
    * Copy constructor and copy assignment operator can throw std::runtime_error if OpenSSL serialization fails.
    */
+  ~ECDSAsecp256k1PrivateKey() override = default;
```
alittley, line 51: Do we actually need to declare that we are using the default destructor on all of these classes? Won't it use the default one anyway if you just say nothing?

alittley (approved):

alittley:
@rwalworth I like the new implementation, since it hides a lot of the nastiness in just that 1 file.

Did you test this new implementation for memory leaks? It would be neat if that could be done in a unit test, not sure how possible that is

sdk/main/include/impl/OpenSSLObjectWrapper.h
rwalworth: I suppose not, removed

sdk/main/include/impl/OpenSSLObjectWrapper.h
rwalworth: Fixed

sdk/main/include/impl/OpenSSLObjectWrapper.h
rwalworth: Fixed

sdk/main/include/impl/OpenSSLObjectWrapper.h
rwalworth: Fixed

sdk/main/include/impl/OpenSSLObjectWrapper.h
rwalworth: Fixed

sdk/main/include/impl/OpenSSLObjectWrapper.h
rwalworth: Fixed

sdk/main/include/impl/OpenSSLObjectWrapper.h
rwalworth: Fixed

sdk/main/include/ECDSAsecp256k1PrivateKey.h
rwalworth: It's always been my impression that it's good practice to, if you have to define one of copy constructor, copy assignment operator, move constructor, move assignment operator, or destructor, then you should define them all, even if they're `default`. With that said, removing this line doesn't have any effect and all tests/examples still pass with no leaks (even tried some stuff that I thought could cause some undefined behavior but it worked fine) so I'm fine removing it if you would like

rwalworth:
> @rwalworth I like the new implementation, since it hides a lot of the nastiness in just that 1 file.
> 
> Did you test this new implementation for memory leaks? It would be neat if that could be done in a unit test, not sure how possible that is

Yes I tested for leaks here and we're still good. I don't believe google test has the capabilities to test for memory leaks as part of unit tests unfortunately, but it's pretty simply to run the whole test suite and see the memory leak info with the `leaks` tool.

sdk/main/include/ECDSAsecp256k1PrivateKey.h
rwalworth: I'm going to merge this as is for now and can come back to it at some point. I was planning on reviewing/updating the implementation of these classes at some point anyways for a couple reasons but primarily to see if there's a way to get rid of these functions. Ideally none of these functions would be needed, but since we want to allow copying and moving of keys (I would think) and this class has `std::unique_ptr` members, they need to be defined.

PR #194, opened 2023-02-06
Conform differences in `verifySignature` for `ECDSAsecp256k1PublicKey` and `ED25519PublicKey`

### LINKED ISSUE #193: ECDSA and Ed25519 sig verification differences
`ECDSAsecp256k1PublicKey::verifySignature` does this:

```
  // any value other than 0 or 1 means an error occurred
  if (verificationResult != 0 && verificationResult != 1)
  {
    throw std::runtime_error(internal::OpenSSLHasher::getOpenSSLErrorMessage("EVP_DigestVerify"));
  }

  return verificationResult == 1;
```

whereas `ED25519PublicKey::verifySignature` does this:

```
  if (verificationResult <= 0)
  {
    std::string message = "Failed to verify signature with code [" + std::to_string(verificationResult) + ']';
    throw std::runtime_error(message.c_str());
  }

  return verificationResult == 1;
```

I think this is a mistake, and the `ECDSA` version is correct. As per the [OpenSSL doc](https://www.openssl.org/docs/man3.0/man3/EVP_DigestVerify.html), 

> EVP_DigestVerifyFinal() and EVP_DigestVerify() return 1 for success; any other value indicates failure. A return value of zero indicates that the signature did not verify successfully (that is, tbs did not match the original data or the signature had an invalid form), while other values indicate a more serious error (and sometimes also indicate an invalid signature form).

IMO it makes sense to *just* return false if verification fails without a more serious error.

rwalworth:
**Description**:
This PR fixes a difference in the `verifySignature` function for our derived `PublicKey` classes, where `false` should be returned if the signature was verified to not be correct, as opposed to throwing an error. Tests were updated to reflect this error as well.

**Related issue(s)**:

Fixes #193 

### DIFF
--- a/sdk/main/include/ECDSAsecp256k1PublicKey.h
+++ b/sdk/main/include/ECDSAsecp256k1PublicKey.h
@@ -96,7 +96,7 @@ public:
    *
    * @param signatureBytes The byte vector representing the signature.
    * @param signedBytes    The bytes which were purportedly signed to create the signature.
-   * @return \c TRUE if the signature is valid, otherwise \c FALSE.
+   * @return \c TRUE if the signature is valid for this ECDSAsecp256k1PublicKey's private key, otherwise \c FALSE.
    * @throws std::runtime_error If OpenSSL signature verification experiences an error.
    */
   [[nodiscard]] bool verifySignature(const std::vector<unsigned char>& signatureBytes,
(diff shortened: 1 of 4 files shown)


alittley (approved):

deyanzz (approved):
LGTM

PR #196, opened 2023-02-06
Add missed wrapper objects

### LINKED ISSUE #173: Replace raw OpenSSL pointer handling with `std::unique_ptr` implementation with custom deleter

rwalworth:
**Description**:
This PR adds some missed OpenSSL wrapper objects that were missed in #191. With that, it also merges the `BigNumber` class into a new `OpenSSL_BIGNUM` class.

**Related issue(s)**:

Fixes #173 

**Checklist**


deyanzz (approved):
LGTM

alittley (approved):

PR #198, opened 2023-02-07
Create more descriptive exceptions

### LINKED ISSUE #170: Create more descriptive exceptions
### Problem

Most exceptions being thrown are simply `std::invalid_argument` and `std::runtime_error`. I believe `std::invalid_argument` is descriptive enough, but we should look to replace `std::runtime_error` with our own custom exceptions.

### Solution

Possibly create an `exceptions/` folder in `sdk/main` that contains a few more descriptive exceptions. Here are some examples I can think of off the top of my head:
 - `OpenSSLException`: handles exceptions that are thrown when OpenSSL fails to accomplish something.
 - `UninitializedException`: handles exceptions that are thrown when a user attempts to use an object/variable that is uninitialized.
 - `PrecheckException`: handles exceptions that are thrown when the precheck for a request fails.

This list is not exhaustive, but would be a good start.

### Alternatives

N/A

rwalworth:
**Description**:
This PR updates exceptions to be more specific. `std::invalid_argument` is fine, but `std::runtime_error` is too broad, and all of these were replaced with new exceptions. This also allows users to be able to catch less exceptions.

Some tests were also added to catch these new exceptions.

**Related issue(s)**:

Fixes #170 

**Checklist**


deyanzz (approved):
LGTM. I double-checked all of the modified tests locally on my MacBook - every test passed successfully!

PR #206, opened 2023-02-17
Add token support to `TransferTransaction`

### LINKED ISSUE #199: Add `TransferTransaction` support for custom tokens
### Problem

As an application developer, I need to use the `TokenTransferTransaction` to transfer NFTs between owners in the application I am developing. 

### Solution

Add support for the `TokenTransferTransaction`.

### Alternatives

_No response_

rwalworth:
**Description**:
This PR adds support to transfer tokens with `TransferTransaction`, as well as needed accompanying classes.

**Related issue(s)**:

Fixes #199 

**Notes for reviewer**:
Functionality from the Java SDK will be used to create a token and associate the token with certain accounts to test this (since these capabilities are required and haven't been developed in the C++ SDK yet).

**Checklist**


rwalworth:
This has been tested by creating and associating tokens and accounts using the Java SDK, then taking those account and token IDs and inputting them into the `TransferTokensExample` to test functionality.

sdk/main/src/HbarTransfer.cc
```diff
@@ -17,16 +17,16 @@
  * limitations under the License.
  *
  */
-#include "Transfer.h"
```
deyanzz, line 20: I'm wondering which one is the actual version and naming for this class - in the [Java SDK](https://github.com/hashgraph/hedera-sdk-java) it's still called [Transfer](https://github.com/hashgraph/hedera-sdk-java/blob/develop/sdk/src/main/java/com/hedera/hashgraph/sdk/Transfer.java), while in [Hedera SDK Reference](https://github.com/hashgraph/hedera-sdk-reference) class `Transfer` is missing.
Shouldn't we rename it to `HbarTransfer` in the Java SDK too?

deyanzz (approved):
LGTM. Tested locally and works fine.

sdk/main/src/HbarTransfer.cc
rwalworth: This is an internal utility class so I'm not sure it matters. I just felt like `Transfer` was too generic now, as this PR added `TokenTransfer` and `TokenNftTransfer`.

PR #209, opened 2023-02-22
Implement `Transaction::fromBytes()`

### LINKED ISSUE #205: Add `Transaction::fromBytes()` function
`Transaction`s should be able to be created from raw bytes. This functionality should be added.

rwalworth:
**Description**:
This PR adds a `Transaction::fromBytes()` function. Unlike what is stated in the [Hedera SDK Reference](https://github.com/hashgraph/hedera-sdk-reference/blob/main/reference/core/Transaction.md#frombytes--data--bytes--transaction) for this function, a clean return of a `Transaction` object is not possible since the return type _must_ be known at compile time. `Transaction` is unable to be returned because it is a template and the template parameter is the created derived class (see [Curiously Recurring Template Pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern)).

To avert this problem, a `std::pair` is returned, which contains a `std::variant` and an `index` into the `std::variant`. While a `switch` statement will be necessary to determine and create the actual derived `Transaction` type, it is the only way to get around the language constricting issue of needing to know the return type at compile time.

**Related issue(s)**:

Fixes #205 

**Checklist**

### DIFF
--- a/sdk/main/include/AccountCreateTransaction.h
+++ b/sdk/main/include/AccountCreateTransaction.h
@@ -67,8 +67,18 @@ namespace Hedera
 class AccountCreateTransaction : public Transaction<AccountCreateTransaction>
 {
 public:
+  /**
+   * Default constructor. Sets the maximum transaction fee to 5 Hbars.
+   */
   AccountCreateTransaction();
-  ~AccountCreateTransaction() override = default;
+
+  /**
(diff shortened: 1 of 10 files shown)


deyanzz (approved):
LGTM. Tested locally and works fine.

PR #210, opened 2023-02-23
Create `OpenSSLUtils` namespace and move all OpenSSL utility classes/functions into it

### LINKED ISSUE #142: Combine OpenSSL utilities into single file
I think it's overkill to have the OpenSSL utilities spread over so many files. They can all be combined into a single utils file

rwalworth:
**Description**:
This PR consolidates the OpenSSL helper functions and classes into a single file.

**Related issue(s)**:

Fixes #142 

**Checklist**


sdk/main/src/impl/OpenSSLUtils.cc
```diff
@@ -17,14 +17,17 @@
  * limitations under the License.
  *
  */
-#include "impl/OpenSSLObjectWrapper.h"
+#include "impl/OpenSSLUtils.h"
 #include "exceptions/OpenSSLException.h"
 #include "impl/HexConverter.h"
-#include "impl/OpenSSLHasher.h"
```
alittley, line 23: Something to consider with this class - it might be worth removing the `OpenSSL` prefix from classes and method names, e.g. `OpenSSLUtils::getErrorMessage` instead of `OpenSSLUtils::getOpenSSLErrorMessage`, `EVP_PKEY` instead of `OpenSSL_EVP_PKEY`, etc.

Since you always fully qualify these when using them anyway, it could really reduce the quantity of text needed to use these objects, without sacrificing clarity. Just a thought

alittley (approved):
LGTM

sdk/main/src/impl/OpenSSLUtils.cc
rwalworth: Very true, I'll remove the extra `OpenSSL` prefix

rwalworth: Fixed

PR #227, opened 2023-03-17
Switch `std::vector<unsigned char>` to `std::vector<std::byte>`

### LINKED ISSUE #213: Look into replacing instances of `unsigned char` with `std::byte`
Reading online seems that `std::byte` should be the standard going forward for any byte-type manipulation. There should be an effort to replacing all `unsigned char` instances that are used for byte processing with `std::byte`.

rwalworth:
**Description**:
This PR switches `std::vector<unsigned char>` usage to `std::vector<std::byte>`. There are a couple benefits to this. First, the `std::byte` provides greater type safety and doesn't allow accidental operations and/or conversions to be performed on them. This makes it a bit more verbose but that verbosity can be encapsulated in helper functions. Also, `std::byte` is more direct with its naming, in that the user knows they should be viewing the unit as a "byte" and not a "char".

**Related issue(s)**:

Fixes #213 

**Checklist**

### DIFF
--- a/sdk/main/include/Client.h
+++ b/sdk/main/include/Client.h
@@ -88,7 +88,7 @@ public:
    * @return The bytes with the signature of this Client's operator appended.
    * @throws UninitializedException If this Client's operator has not yet been set.
    */
-  [[nodiscard]] std::vector<unsigned char> sign(const std::vector<unsigned char>& bytes) const;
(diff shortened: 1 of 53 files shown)


deyanzz (approved):
LGTM

Issue #265, opened 2023-04-06
Failing steps in the PR checks workflow


deyanzz:
### Description

Currently, in the `main` branch, some of the steps in the PR check actions are disabled. 
Names of all disabled steps:

- `Install Local Node`
- `Start the local node`
- `Start CTest suite (Debug)`
- `Start CTest suite (Release)`
- `Stop the local node`

First step from the list (`Install Local Node`) is failing withe the following error:

> Waiting for the containers at 127.0.0.1:5600, retrying in 0.1 seconds...
Error: connect ECONNREFUSED 127.0.0.1:5600
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16) {
  errno: -111,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5600
}

Example of failed PR check - [Job #8174398369](https://github.com/hashgraph/hedera-sdk-cpp/actions/runs/4622119534/jobs/8174398369)

### Steps to reproduce

Just remove the comment # for the following steps:

- `Install Local Node`
- `Start the local node`
- `Start CTest suite (Debug)`
- `Start CTest suite (Release)`
- `Stop the local node`

and initiate a new PR check action.

### Additional context

The following [GitHub Actions extension](https://github.blog/2023-03-28-announcing-the-github-actions-extension-for-vs-code/) for VS Code could be used.  
Tutorial how to run manually a given workflow - [Manually running a workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow?tool=cli).

### Hedera network

other

### Version

v0.5.0

### Operating system

Linux

Issue #164, opened 2023-01-18
Test for create and execute requests to send to a node


deyanzz:
Invoking of method `execute()` should be tested using a [Hedera Local Node](https://github.com/hashgraph/hedera-local-node) for the following classes:

- `AccountBalanceQueryIntegrationTest`
- `AccountCreateTransactionIntegrationTest`
- `TransactionReceiptQueryIntegrationTest`
- `TransactionRecordQueryIntegrationTest`
- `TransferTransactionIntegrationTest`

See the [README.md](https://github.com/hashgraph/hedera-local-node/blob/main/README.md) for the requirement how to install the local node.

deyanzz:
Current implementation is not working well. Will be back to finish it once a new PR is opened for [Issue 163](https://github.com/hashgraph/hedera-sdk-cpp/issues/163).

deyanzz:
To be able to make a working test for `AccountCreateTransaction`, first we need to fix [Issue #192](https://github.com/hashgraph/hedera-sdk-cpp/issues/192).

deyanzz:
This issue is currently blocked by missing capabilities in `Client` - see [Issue #201](https://github.com/hashgraph/hedera-sdk-cpp/issues/201).
Once [Issue #201](https://github.com/hashgraph/hedera-sdk-cpp/issues/201) is merged into main branch the integration tests could be finished.

deyanzz:
This issue is blocked by [Issue #257](https://github.com/hashgraph/hedera-sdk-cpp/issues/257).

deyanzz:
Continue working on this issue as [issue #257](https://github.com/hashgraph/hedera-sdk-cpp/issues/257) and [issue #201](https://github.com/hashgraph/hedera-sdk-cpp/issues/201) were merged into `main` branch.

PR #267, opened 2023-04-11
Failing steps in the PR checks workflow (PR)

deyanzz:
**Description**:

This PR tries to fix an issue, which makes the PR checks to fail with the following error:

> Waiting for the containers at 127.0.0.1:5600, retrying in 0.1 seconds...
Error: connect ECONNREFUSED 127.0.0.1:5600
at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16) {
errno: -111,
code: 'ECONNREFUSED',
syscall: 'connect',
address: '127.0.0.1',
port: 5600
}

Link to a failed [PR Checks](https://github.com/hashgraph/hedera-sdk-cpp/actions/runs/4687757526/jobs/8307440898) flow.
If the flow is unable to run a [Hedera Local Node](https://github.com/hashgraph/hedera-local-node) then no integration tests could be executed during the `PR Checks`.

**Related issue(s)**: [Test for create and execute requests to send to a node - #164](https://github.com/hashgraph/hedera-sdk-cpp/issues/164) 

**Fixes:** [Failing steps in the PR checks workflow - #265](https://github.com/hashgraph/hedera-sdk-cpp/issues/265)

**Notes for reviewer**:
I tried several approaches to discover why this port is not reachable. 

1) First, I tried to execute the command 
`lsof -Pni | grep "5600"`
but I ended up with an error message: `lsof: command not found` [PR Checks attempt](https://github.com/hashgraph/hedera-sdk-cpp/actions/runs/4687194312/jobs/8306179328) 
When I added an installation step in the Action flow, it ended up with another error message:
`Process completed with exit code 1.`
[PR Checks attempt](https://github.com/hashgraph/hedera-sdk-cpp/actions/runs/4687239900/jobs/8306282071)
Which leads me to the conclusion `lsof` couldn't be used in this flow.

2) After that I tried to change the OS which runs the flow - I commented the following lines:

> runs-on:
        - self-hosted
        - "${{ matrix.os }}"
        - xlarge
        - ephemeral

I changed it to run-on `ubuntu-latest`:

> runs-on: ubuntu-latest

and the [PR Checks](https://github.com/hashgraph/hedera-sdk-cpp/actions/runs/4740752240/jobs/8416945767) flow passed successfully, but it took enormous amount of time - `1h 27m 36s`

So I decided to open this PR with the working solution of the issue but also to initiate a discussion about how to improve it - how to reduce the run time for the whole flow.

3) My latest attempt is in this [PR Checks](https://github.com/hashgraph/hedera-sdk-cpp/actions/runs/4744241977/jobs/8424884898) flow, where I commented all `Release` related steps to reduce run time of the flow.

**Checklist**

### DIFF
--- a/.github/workflows/zxc-build-library.yaml
+++ b/.github/workflows/zxc-build-library.yaml
@@ -1,6 +1,12 @@
 name: "ZXC: Build Library"
 on:
-  workflow_call:
+  #workflow_call:
+  pull_request:
+      branches:
+          - main
+  push:
+      branches:
+          - main
 
 defaults:
   run:
@@ -10,41 +16,33 @@ permissions:
   contents: read
 
 jobs:
-  build:
-    name: Build
-    runs-on:
-      - self-hosted
-      - "${{ matrix.os }}"
-      - xlarge
-      - ephemeral
-    strategy:
-      matrix:
-        include:
-          - os: Linux
-            preset: linux-x64
-    #          - os: macos-12
-    #            preset: macos-x64
-    #          - os: windows-latest
-    #            preset: windows-x64
-    steps:
-      - name: Checkout Code
-        uses: actions/checkout@v3
-        with:
-          submodules: true
+    build:
+      name: Build
+      runs-on:
+        - self-hosted
+        - "${{ matrix.os }}"
+        - xlarge
+        - ephemeral
+      strategy:
+        matrix:
+          include:
+            - os: Linux
+              preset: linux-x64
+            - os: macos-12
+              preset: macos-x64
+            - os: windows-latest
+              preset: windows-x64
+      steps:
+          - name: Checkout Code
+            uses: actions/checkout@v3
+            with:
+              submodules: true
 
-      - name: Install Linux Prerequisites
-        if: ${{ runner.os == 'Linux' }}
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y pkg-config libc6-dbg libgtest-dev build-essential
-
-      - name: Use Node.js 18
-        uses: actions/setup-node@v3
-        with:
-          node-version: 18
-
-      - name: Install Local Node
-        run: sudo npm install -g @hashgraph/hedera-local
+          - name: Install Linux Prerequisites
+            if: ${{ runner.os == 'Linux' }}
+            run: |
+              sudo apt-get update
+              sudo apt-get install -y pkg-config libc6-dbg libgtest-dev build-essential
 
       # - name: Install lsof
       #   run: sudo apt-get install lsof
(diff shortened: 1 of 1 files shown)


.github/workflows/zxc-build-library.yaml
```diff
@@ -10,94 +10,114 @@ permissions:
...
+                  curl \
+                  gnupg \
+                  lsb-release
+              sudo mkdir -p /etc/apt/keyrings
+              curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+              echo \
+                "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
+                $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+              sudo apt-get update
+              sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
```
nathanklick, line 51: This step needs to be restricted to Linux only using an `if` clause or needs to be replaced with a universal step such as https://github.com/marketplace/actions/docker-setup-docker which provides WIndows, Linux, and MacOS support. 

I strongly suggest using the universal step from crazymax because we must ultimately support Windows and MacOS as well in these workflows.

.github/workflows/zxc-build-library.yaml
```diff
@@ -10,94 +10,114 @@ permissions:
...
+              configurePreset: ${{ matrix.preset }}-debug
+              buildPreset: ${{ matrix.preset }}-debug
+        
+          # - name: CMake Build (Release)
+          #   uses: lukka/run-cmake@v10
+          #   env:
+          #     VCPKG_BINARY_SOURCES: clear
+          #   with:
+          #     configurePreset: ${{ matrix.preset }}-release
+          #     buildPreset: ${{ matrix.preset }}-release
```
nathanklick, line 98: We must build both release and debug of the libraries. Building release only takes a fraction of the time compared to the debug build.

.github/workflows/zxc-build-library.yaml
```diff
@@ -10,94 +10,114 @@ permissions:
...
+          # - name: List Directory Structure
+          #     if: ${{ contains(runner.os, 'Windows') && always() }}
+          #     shell: pwsh
+          #     run: tree /F build
+        
+          - name: Start CTest suite (Debug)
+            run: ctest -C Debug --test-dir build/${{ matrix.preset }}-debug --output-on-failure
+        
+          # - name: Start CTest suite (Release)
+          #   run: ctest -C Debug --test-dir build/${{ matrix.preset }}-release --output-on-failure
```
nathanklick, line 109: We must build both release and debug of the libraries. Building release only takes a fraction of the time compared to the debug build.

nathanklick (changes requested):

.github/workflows/zxc-build-library.yaml
isavov: It seems the version installed via the `crazy-max/ghaction-setup-docker` conflicts with the one that's already globally installed in the runner image. The options that I see are:
* try to develop some workaround to kill and remove the preinstalled version
* update the globally installed version in a `if:` step
* update the images used for the runner with a new version of `docker/docker compose`

.github/workflows/zxc-build-library.yaml
deyanzz: Release builds has been activated again with 29b615968f94f078ec8f9dced266e0418078085f.
I commented them only to reduce the run time during my attempts to fix the issue.
I know how important is this step and my plan was to remove the comments at the end, but I forgot.

.github/workflows/zxc-build-library.yaml
deyanzz: Running `CTest` suite has been activated again with 29b615968f94f078ec8f9dced266e0418078085f.
I commented them only to reduce the run time during my attempts to fix the issue - same reason as for the `CMake` builds.

.github/workflows/zxc-build-library.yaml
nathanklick: Have you tried specifying a different version such as `20.10.24` to the crazymax plugin instead of the defaulr (`latest`)?

.github/workflows/zxc-build-library.yaml
```diff
@@ -10,94 +10,112 @@ permissions:
...
+              sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
+            if: ${{ runner.os == 'Linux' }}
+              
+          - name: Use Node.js 18
+            uses: actions/setup-node@v3
+            with:
+              node-version: 18
+        
+          - name: Install Local Node
+            run: sudo npm install -g @hashgraph/hedera-local
```
nathanklick, line 60:
```suggestion
            run: sudo npm install -g @hashgraph/hedera-local
            if: ${{ runner.os == 'Linux' }}
```

nathanklick (approved):

PR #319, opened 2023-05-24
`RLPValue` library throws when a value is empty

### LINKED ISSUE #314: `RLPValue` library throws when a value is empty
### Description

The `EthereumTransactionData` unit tests are not passing on Windows. This is due to the `RLPValue` library having a bug on Windows, where converting an `RLPBuffer` to a `std::string` doesn't check to see if the the data vector is empty first. Manually adding a check in there allows all tests to pass, but this is an undesirable solution. Currently the library offers no way to make this check. A new library will need to be found or we will have to write one ourselves.

### Steps to reproduce

1. Build the `hedera-sdk-cpp-unit-tests.exe` on Windows.
2. Run it.
3. Observe failing tests.

### Additional context

_No response_

### Hedera network

other

### Version

v0.8.0-dev

### Operating system

Windows

rwalworth:
**Description**:
This PR fixes a bug that was caused by a bug in the RLPValue library. Basically, the RLPValue library provided no way to check if a value was empty or not without trying to index into the first element, which throws on Windows builds. This PR removed that library and rewrites a condensed version of it that fixes this issue.

**Related issue(s)**:

Fixes #314 

**Checklist**

### DIFF
--- a/sdk/main/CMakeLists.txt
+++ b/sdk/main/CMakeLists.txt
@@ -3,20 +3,6 @@ FetchContent_Declare(valuable
         GIT_REPOSITORY https://github.com/LoopPerfect/valuable.git)
 FetchContent_MakeAvailable(valuable)
 
-FetchContent_Declare(rlpvalue
-        GIT_REPOSITORY https://github.com/bloq/rlpvalue.git
-        GIT_TAG master)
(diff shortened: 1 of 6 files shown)


deyanzz (approved):
LGTM. Local tests passes successfully.

PR #325, opened 2023-05-25
Attempting to work with `std::vector<const Key*>` was causing crashing.

### LINKED ISSUE #320: Attempting to work with `KeyList` objects made from `std::vector<const Key*>` results in crashing
### Description

Attempting to work with a `KeyList` object that was created from a `std::vector<const Key*>` container results in the seg faults. This is probably caused by a bug in `ValuePtr` somewhere, will require some investigation.

### Steps to reproduce

1. Create a `std::vector<const Key*>` and construct a `KeyList` from it (using `of`).
2. Observe crashing.

### Additional context

_No response_

### Hedera network

other

### Version

v0.9.0-dev

### Operating system

macOS

rwalworth:
**Description**:
This PR addresses an issue that was found when trying to write tests for `FileCreateTransaction` and `FileUpdateTransaction`. When attempting to set the keys in aforementioned transactions via a `std::vector<const Key*>` container, the tests were crashing. This actually did not end up being a bug in the processing of the keys, but the fact that they weren't being setup properly. When creating a `std::vector<const Key*>` with something like `const std::vector<const Key*> mTestKeyVector = {ED25519PrivateKey::fromString(...).get(), ...};`, the created `std::unique_ptr<ED25519PrivateKey>` was going out of scope and the `mTestKeyVector` just contained a bunch of deleted pointers. This was done correctly and more tests were added to `FileCreateTransaction` and `FileUpdateTransaction`.

In addition, I changed a couple of other things:
 - I changed functions with input parameter `std::vector<const Key*>` to `std::vector<Key*>`. There's no reason to enforce the `const` and those functions will just create and keep a copy anyways.
 - In `KeyList::of`, I used `KeyCloner` to clone the input keys instead of dereferencing, since you can't dereference an abstract class anyways. Honestly, not sure how this worked in the first place, but at least this change keeps processing consistent.

**Related issue(s)**:

Fixes #320 

**Checklist**


deyanzz (approved):
LGTM. All tests passed successfully.

Issue #284, opened 2023-04-26
Integration tests - common logic to setup the local node


deyanzz:
### Problem

Currently, the following integration tests:

- `AccountAllowanceApproveTransactionIntegrationTest`
- `AccountBalanceQueryIntegrationTest`
- `AccountCreateTransactionIntegrationTest`
- `AccountDeleteTransactionIntegrationTest`
- `AccountInfoQueryIntegrationTest`
- `AccountRecordsQueryIntegrationTest`
- `AccountUpdateTransactionIntegrationTest`
- `ContractBytecodeQueryIntegrationTest`
- `ContractCallQueryIntegrationTest`
- `ContractCreateTransactionIntegrationTest`
- `ContractDeleteTransactionIntegrationTest`
- `ContractExecuteTransactionIntegrationTest`
- `ContractInfoQueryIntegrationTest`
- `ContractUpdateTransactionIntegrationTest`

 are using the same logic to connect to a local node:

> void SetUp() override
  {
    const auto accountId = AccountId::fromString("0.0.3");
    const std::string_view accountIdStr = "0.0.3";
    const std::string_view networkTag = "network";
    const std::string_view operatorTag = "operator";
    const std::string_view accountIdTag = "accountId";
    const std::string_view privateKeyTag = "privateKey";
    const std::string testPathToJSON = std::filesystem::current_path() / "local_node.json";
    const std::unique_ptr<PrivateKey> testPrivateKey = ED25519PrivateKey::generatePrivateKey();
    const std::shared_ptr<PublicKey> testPublicKey = testPrivateKey->getPublicKey();

> AccountId operatorAccountId;
    std::string operatorAccountPrivateKey;
    std::ifstream testInputFile(testPathToJSON, std::ios::in);
    std::string nodeAddressString;
    json jsonData = json::parse(testInputFile);
    if (jsonData[networkTag][accountIdStr].is_string())
    {
      nodeAddressString = jsonData[networkTag][accountIdStr];
    }
    if (jsonData[operatorTag][accountIdTag].is_string() && jsonData[operatorTag][privateKeyTag].is_string())
    {
      std::string operatorAccountIdStr = jsonData[operatorTag][accountIdTag];
      operatorAccountId = AccountId::fromString(operatorAccountIdStr);
      operatorAccountPrivateKey = jsonData[operatorTag][privateKeyTag];
    }
    testInputFile.close();
    std::unordered_map<std::string, AccountId> networkMap;
[...]

PR #339, opened 2023-06-06
Integration tests - common logic to setup the local node - Part 1 (PR)

deyanzz:
**Description**:
This PR introduces a new base class `BaseIntegrationTest` and some refactoring around it. All of the common logic for connecting to a [Hedera Local Node](https://github.com/hashgraph/hedera-local-node) has been moved into `BaseIntegrationTest`.

`BaseIntegrationTest` has been added as a super class for the following integration tests:
- `AccountAllowanceApproveTransactionIntegrationTest`
- `AccountBalanceQueryIntegrationTest`
- `AccountCreateTransactionIntegrationTest`
- `AccountDeleteTransactionIntegrationTest`
- `AccountInfoQueryIntegrationTest`
- `AccountRecordsQueryIntegrationTest`
- `AccountUpdateTransactionIntegrationTest`
- `ContractBytecodeQueryIntegrationTest`
- `ContractCallQueryIntegrationTest`
- `ContractCreateTransactionIntegrationTest`
- `ContractDeleteTransactionIntegrationTest`
- `ContractExecuteTransactionIntegrationTest`
- `ContractInfoQueryIntegrationTest`
- `ContractUpdateTransactionIntegrationTest`

**Related issue(s)**: 
This PR is a mandatory part to be able to start working on issue [#340 - Integration tests - common logic to setup the local node - Part 2](https://github.com/hashgraph/hedera-sdk-cpp/issues/340)

**Fixes:** [#284](https://github.com/hashgraph/hedera-sdk-cpp/issues/284)

**Notes for reviewer**:

- Second portion of refactoring will be introduced within  issue [#340](https://github.com/hashgraph/hedera-sdk-cpp/issues/340).
- Suggested improvement - method `getTestSmartContractBytecode()` from class `ContractBytecodeQueryIntegrationTest` could also be moved in `BaseIntegrationTest`. 
(EDIT: Already done in 2d9ab452516fd6f9dfff95dc8ad828bf40492bf6).

**Checklist**


sdk/tests/integration/BaseIntegrationTest.cc
```diff
@@ -0,0 +1,81 @@
...
+ */
+#include "BaseIntegrationTest.h"
+#include "AccountId.h"
+#include "Client.h"
+#include "ECDSAsecp256k1PrivateKey.h"
+#include "ED25519PrivateKey.h"
+
+#include <filesystem>
+#include <fstream>
+#include <gtest/gtest.h>
```
rwalworth, line 28: This is an unused include in this file, it can be removed.

sdk/tests/integration/BaseIntegrationTest.cc
```diff
@@ -0,0 +1,81 @@
...
+#include <nlohmann/json.hpp>
+#include <string>
+#include <string_view>
+#include <unordered_map>
+
+using json = nlohmann::json;
+using namespace std;
```
rwalworth, line 35: All `std` objects in this file still use the namespace scoping (i.e. `std::string_view`). I'd recommend removing this.

sdk/tests/integration/BaseIntegrationTest.cc
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Hedera
+{
+//-----
+void BaseIntegrationTest::SetUp()
+{
+  const auto accountId = AccountId::fromString("0.0.3");
+  const std::string_view accountIdStr = "0.0.3";
```
rwalworth, line 43: I don't think we want these hardcoded values moving forward. I think it'd be better for use to have processing that would loop through the "network" JSON value and build the network from there. If a configuration was fed into the integration tests that didn't use `0.0.3`, this would break.

sdk/tests/integration/BaseIntegrationTest.cc
```diff
@@ -0,0 +1,81 @@
...
+  const std::string_view networkTag = "network";
+  const std::string_view operatorTag = "operator";
+  const std::string_view accountIdTag = "accountId";
+  const std::string_view privateKeyTag = "privateKey";
+
+  const std::string testPathToJSON = (std::filesystem::current_path() / "local_node.json").string();
+  const std::unique_ptr<PrivateKey> testPrivateKey = ED25519PrivateKey::generatePrivateKey();
+  const std::shared_ptr<PublicKey> testPublicKey = testPrivateKey->getPublicKey();
```
rwalworth, line 51: These `testPrivateKey` and `testPublicKey` values don't look used, they can be removed.

sdk/tests/integration/BaseIntegrationTest.h
```diff
@@ -0,0 +1,43 @@
...
+#define HEDERA_SDK_CPP_BASE_INTEGRATION_TEST_H_
+
+#include "Client.h"
+
+#include <gtest/gtest.h>
+
+namespace Hedera
+{
+
+class BaseIntegrationTest : public testing::Test
```
rwalworth, line 30: The JSON file is going to get read for each instance of a `BaseIntegrationTest`, and establishing a network map and client for each, when really it should only need to happen once. Is there a way to rework this so that we only do this once?

rwalworth (changes requested):
There are a couple small things I think could be changed, and I think this should be reworked so that the client is constructed only once. I'm thinking making just a separate `IntegrationTestEnv` class of some sort that has an `init()` function that can be called in each test's `SetUp()` that checks if its already initialized with the JSON data and reads the JSON data if it's not initialized. This makes sure that the client and network map are initialized for each test, is independent of test order, and only reads in the JSON file once.

sdk/tests/integration/BaseIntegrationTest.cc
deyanzz: Removed in 451faabb11a8b67c1bb998dcee410b4de2c284af.

sdk/tests/integration/BaseIntegrationTest.cc
deyanzz: Removed in 451faabb11a8b67c1bb998dcee410b4de2c284af.

sdk/tests/integration/BaseIntegrationTest.cc
deyanzz: Agreed. This part has been reworked in 83259f5adc608fae6d8102294065ded2dacb5d3d.

sdk/tests/integration/BaseIntegrationTest.cc
deyanzz: I brought back the `using namespace std;` in the beginning of the file bur removed all of the `std::` namespace scoping.

sdk/tests/integration/BaseIntegrationTest.h
deyanzz: I agree there's a better way to handle JSON file loading - we can create a Singleton class named `IntegrationTestEnv ` - see https://github.com/hashgraph/hedera-sdk-cpp/issues/349 for more details.

deyanzz:
> There are a couple small things I think could be changed, and I think this should be reworked so that the client is constructed only once. I'm thinking making just a separate `IntegrationTestEnv` class of some sort that has an `init()` function that can be called in each test's `SetUp()` that checks if its already initialized with the JSON data and reads the JSON data if it's not initialized. This makes sure that the client and network map are initialized for each test, is independent of test order, and only reads in the JSON file once.

I totally agree with you - using a Singleton class `IntegrationTestEnv` will encapsulate the JSON file loading in one place only. Also the parsing could be done only once. I just created this new [issue #349](https://github.com/hashgraph/hedera-sdk-cpp/issues/349) to track this improvement.

Also in 2d9ab452516fd6f9dfff95dc8ad828bf40492bf6 I moved method `getTestSmartContractBytecode()` in the base class `BaseIntegrationTest`. This also reduce code duplication in some of integration tests related to Smart Contracts.

rwalworth (approved):

PR #454, opened 2023-08-15
Implement `ScheduleCreateTransaction`

### LINKED ISSUE #451: Implement `ScheduleCreateTransaction`

rwalworth:
**Description**:
This PR implements the APIs associated with `ScheduleCreateTransaction`, which allows a user to schedule a transaction to execute at some later point (either at expiration or when all signatures on the transaction are received). This PR also implements and rearranges a couple of other things. Notable changes:

 - `Transaction::freezeWith()` now passes in a `const Client*` instead of a `const Client&`. This allows `Transaction`s to be frozen without the need of a `Client` (the `Client*` would equal `nullptr`).
 - Add an `addToBody(proto::TransactionBody&)` virtual function to `Transaction`s, which allows each derived `Transaction` to place its data into the corresponding `proto::TransactionBody` data field.
 - `ChunkedTransaction`s now have an `onChunk()` function, which allows processing to occur when a chunk is being processed/executed.
 - Add `Transaction::schedule()`, which allows any `Transaction` to put itself in a `ScheduleCreateTransaction`.
 - Add `TransactionId::withValidStart()`, which allows a `TransactionId` to be set with an account ID and a start time. This is needed when generating the `TransactionId`s for chunks in a `ChunkedTransaction`.
 - Add `enum TransactionType` and `class WrappedTransaction`. This is a much less cumbersome solution for unknown `Transaction` types then the previous `std::variant`. These provide APIs to allow `Transaction`s to be used generically given the class hierarchy.
 - Add `UnsupportedOperationException`, which is self-explanatory.
 - Revamp `Transaction::fromBytes()` to use `WrappedTransaction` as a return type.
 - Actually freeze a `Transaction` when executing it.
 - 

**Related issue(s)**:

Fixes #451 

**Checklist**


deyanzz (approved):
LGMT. Good to be merged into `main` branch.

PR #507, opened 2023-09-21
Add `GetAddressBookExample`

rwalworth:
**Description**:
This PR adds all processing associated with `AddressBookQuery`. This includes:
- The `AddressBookQuery` and execution processing.
- The automatic `Client` network update.
- The `GetAddressBookExample`

This PR made `NodeAddressBook`, `NodeAddress`, `Endpoint`, and `IPv4Address` part of the SDKs public API as well.

**Related issue(s)**:

Fixes #504 

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #508, opened 2023-09-22
Update address book `*.pb` files

### LINKED ISSUE #115: Update address book .pb files
The current address books which were sourced from the golang SDK are out of date. The testnet address book, for example, only includes a single endpoint for each node, rather than listing one separately for the TLS and cleartext ports.

For now, I put a check in the `Endpoint` constructor which changes any non TLS ports into the TLS port. This was for testing purposes to be able to develop the TLS capability, and has the effect of making all communications be TLS for now.

Once the API call to fetch the address book is implemented, the response protobuf bytes for each network should be written to the address book files.

Once this is complete, the port swap in `Endpoint` will no longer be necessary. At that time, the user will properly be able to select between TLS being enabled or disabled

rwalworth:
**Description**:
This PR simply updates the SDK's address book files to the most recent versions.

**Related issue(s)**:

Fixes #115 

**Notes for reviewer**:
This PR depends on changes in #507, so that PR should be approved and merged first. 

**Checklist**


deyanzz (approved):
LGTM.

Issue #373, opened 2023-06-14
Tests for min and max backoff time


deyanzz:
### Problem

Currently, the C++ SDK is missing Unit and Integration tests to verify min and max backoff time when connecting to a Node.

### Solution

To add Unit and Integration tests to check Min and Max backoff time.

### Alternatives

N/A

PR #510, opened 2023-10-02
Tests for min and max backoff time (PR)

deyanzz:
**Description**:

This PR adds more integration tests to check the minimum and maximum backoff time configuration for:

- `Client`

**Related issue(s)**:

**Fixes**: https://github.com/hashgraph/hedera-sdk-cpp/issues/373

**Notes for reviewer**:

**Checklist**


rwalworth (approved):
These should probably be a part of unit tests since it's just testing that the values get set in the Client, but they can be moved over at a later time

deyanzz:
Agree. These tests should be part of the Unit tests suite.

PR #518, opened 2023-10-10
Get rid of `ValuePtr` and replace `Key*` instances with `std::shared_ptr<Key>`

### LINKED ISSUE #517: Replace `Key*` with `std::shared_ptr<Key>`
### Problem

Unmanaged passing of `Key` objects can cause confusion about ownership and can lead to crashing. `std::shared_ptr<Key>` should be used to properly pass `Key` objects around and take ownership out of the equation.

### Solution

Use `std::shared_ptr<Key>` and get rid of `ValuePtr<Key>` as making copies is unnecessary if using `std::shared_ptr<Key>`.

### Alternatives

_No response_

rwalworth:
**Description**:
This PR replaces `Key*` instances with `std::shared_ptr<Key>`. `std::shared_ptr<Key>` is safer and doesn't raise the question of ownership when accessing `Key` objects. This will prevent instances where a `Key` goes out of scope and is then attempted to be accessed by a class, which would result in a crash. This also means `ValuePtr` for the most part is deprecated, so it was removed.

**Related issue(s)**:

Fixes #517 

**Checklist**


deyanzz (approved):
LGTM

Issue #546, opened 2023-10-18
Project ```examples/``` runs resulting in errors


gsstoykov:
## Description
Hedera C++ SDK examples failing on startup after successful build on MacOS. 

## Expected Behavior
Normal flow of execution defined in the examples and successful termination.

## Actual Behavior
Primarily ```std::out_of_range``` is thrown on example execution. But has resulted in several other exceptions being thrown.

## Environment
- MacOS
- Apple M2 Pro
- Built with the basic guideline instructions

## Definition of Done
- Analysed and proposed fix
- Implementation resulting in proper example executions

### Steps to reproduce

Example run: 
hedera-sdk-cpp-auto-create-account-transfer-transaction-example [AccountId, PKey]

### Hedera network

testnet

### Version

v0.18.0

### Operating system

macOS

gsstoykov:
Created script for automating the example executions. Error in the example executions was fixed moving to Review.

PR #548, opened 2023-10-18
Project examples/ runs resulting in errors (PR)

gsstoykov:
**Description**:

This PR fixes the following:
- Correctly running the examples on MacOS
- VSCode related files where still committed to the repository

**Related issue(s):**

**Fixes:** https://github.com/hashgraph/hedera-sdk-cpp/issues/546 

**Notes for reviewer**:

The added run_examples script is created with the purpose of having an easier way to identify if there is an overall problem with the example run flows.

**Checklist**


rwalworth (approved):
LGTM

sdk/main/src/impl/Network.cc
```diff
@@ -189,7 +190,8 @@ NodeAddressBook Network::getAddressBookForLedgerId(const LedgerId& ledgerId)
     return {};
   }
 
-  std::ifstream infile(ledgerId.toString() + ".pb", std::ios_base::binary);
+  std::string buildPath = std::filesystem::current_path().string() + "/addressbook/" + ledgerId.toString()+".pb";
```
rwalworth, line 193: This actually breaks running examples from the `build` folder

rwalworth (changes requested):

scripts/run_examples.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+# Iterate through the executables in the directory
+for exe in "$exe_dir"/*; do
+    if [ -x "$exe" ]; then
+        echo "Running $exe with arguments: $arg1 $arg2"
+        "$exe" "$arg1" "$arg2"
+        echo "Finished running $exe"
+    fi
+done
```
rwalworth, line 23: Is there a way to add being able to just run executable? Some of the examples never finish (end with a `while (true) {}` loop) so this won't run through all examples anyways

rwalworth: I might recommend installing this script as well, so anybody who downloads a release can run examples the same way

rwalworth: Also, not all examples run with the same two input arguments (see `ConstructClientExample` and `GeneratePrivateKeyFromMnemonicExample`). Should these maybe be included in a separate script? Or possibly a script made for each example?

gsstoykov: Will move the script topic to a separate discussion and issue.

deyanzz (approved):
LGTM.

Issue #547, opened 2023-10-18
Add HTTP client for querying ```MirrorNode```


gsstoykov:
## Description
While working on https://github.com/hashgraph/hedera-sdk-cpp/issues/498 I discovered that in the C++ SDK there is no way to query the ```MirrorNetwork``` API.

## Use Case
An HTTP client can be invaluable for interacting with the ```MirrorNetwork``` and not only.  A need may arise to query different types of API`s.

## Current Behavior
Not implemented

### Solution

- Agree on proper tooling (ex. cURL)
- Add tooling dependencies to be fetched(ex. via vcpkg)
- Implement a Service to do the proper http queries

Issue #498, opened 2023-09-04
AccountId#toSolidityAddress() does not return expected result when AccountId has a Key Alias


svetoslav-nikol0v:
### Problem

 ```
const createAccountTxRecordWithChildren = await new TransactionRecordQuery()
        .setTransactionId(createAccountTxRecord.transactionId)
        .setIncludeChildren(true)
        .execute(client);
    const multisigAccountId =
        createAccountTxRecordWithChildren?.children[0]?.receipt?.accountId;
    const multisigAccountEvmAddress =
        createAccountTxRecordWithChildren?.children[0]?.evmAddress;
    console.log('multisigAccountId', multisigAccountId.toString());
    console.log('multisigAccountEvmAddress', multisigAccountEvmAddress.toString());
    console.log('toLongZeroEvmAddress(multisigAccountId)', toLongZeroEvmAddress(multisigAccountId));
    console.log('multisigAccountId.toSolidityAddress()', multisigAccountId.toSolidityAddress());
```

```
multisigAccountId 0.0.630405
multisigAccountEvmAddress 6c48bca69c6e7ebcc565f4b79d890eb1fe45e07f
toLongZeroEvmAddress(multisigAccountId) 0000000000000000000000000000000000099e85
multisigAccountId.toSolidityAddress() 0000000000000000000000000000000000099e85
```

### Solution

- In this scenario I expect multisigAccountEvmAddress and multisigAccountId.toSolidityAddress() to be the same - the "regular" EVM address
- And expect only toLongZeroEvmAddress(multisigAccountId) (which is EntityIdHelper.toSolidityAddress([shard, realm, num])) to give me the long zero address

### Alternatives

[Link to issue in JS SDK](https://github.com/hashgraph/hedera-sdk-js/issues/1817).

gsstoykov:
The fix implementation is dependant on having a proper channel from which the ```MirrorNode``` can be queried. I checked the linked issue from the JS SDK which brings the attention to creating a function ```populateAccountEvmAddress```. The function goal is fetching the missing ```EvmAddress``` data from the ```MirrorNode```. In the C++ SDK we can try to get the instance of the ```MirrorNode``` but the get call results in a segmentation fault coming from a call to ```std::unique_lock lock(*getLock());``` this arises concerns for bad synchronization in the internal state of the ```MirrorNode```. Should be further analysed.

gsstoykov:
[...]

PR #555, opened 2023-10-20
Add HTTP Client for querying MirrorNode (PR)

gsstoykov:
**Description**:
A need arised from https://github.com/hashgraph/hedera-sdk-cpp/issues/498 to have a way to query the ```MirrorNode``` in order to fetch ```evmAdress``` field. Turned out there was no ```HttpClient``` being used by the SDK. I decided to use the ```libcurl``` for C++ and added it to project dependencies. Implemented basic functions for invoking RPC methods and HTTP GET/POST. 

**Related issue(s)**:
https://github.com/hashgraph/hedera-sdk-cpp/issues/498
https://github.com/hashgraph/hedera-sdk-cpp/issues/547

**Fixes** https://github.com/hashgraph/hedera-sdk-cpp/issues/547

**Notes for reviewer**:

**Checklist**


sdk/main/CMakeLists.txt
```diff
@@ -152,7 +152,8 @@ add_library(${PROJECT_NAME} STATIC
         src/impl/OpenSSLUtils.cc
         src/impl/RLPItem.cc
         src/impl/TimestampConverter.cc
-        src/impl/Utilities.cc)
+        src/impl/Utilities.cc
+        src/impl/HttpClient.cc)
```
rwalworth, line 156: Small thing, can you keep this list alphabetized?

sdk/main/include/impl/HttpClient.h
```diff
@@ -0,0 +1,58 @@
+#include <curl/curl.h>
```
rwalworth, line 1: Please add file header.

sdk/main/include/impl/HttpClient.h
```diff
@@ -0,0 +1,58 @@
...
+#include <iostream>
+#include <string>
+#include <memory>
+
+class HttpClient
```
rwalworth, line 6: Please include class in `namespace Hedera`.

sdk/main/include/impl/HttpClient.h
```diff
@@ -0,0 +1,58 @@
...
+   *
+   * This constructor initializes the HttpClient and performs global libcurl initialization
+   * using CURL_GLOBAL_DEFAULT.
+   *
+   * @throw std::runtime_error If libcurl initialization fails.
+   */
+  HttpClient();
+
+  HttpClient(const HttpClient&) = delete;            // no need as only one httpClient could be used
+  HttpClient& operator=(const HttpClient&) = delete; // no need as only one httpClient could be used
```
rwalworth, line 20: If only one `HttpClient` can/should be used, should this class be made a singleton?

sdk/main/src/impl/HttpClient.cc
```diff
@@ -0,0 +1,91 @@
+#include "impl/HttpClient.h"
```
rwalworth, line 1: Please include file header.

sdk/main/include/impl/HttpClient.h
```diff
@@ -0,0 +1,58 @@
...
+   *  Can be further extended for supporting other HTTP
+   *  methods or handle more advanced scenarios as needed.
+   * @param url The URL to fetch data from.
+   * @param httpMethod The HTTP method.
+   * @param requestBody The HTTP request body.
+   * @return The fetched data as a string.
+   */
+  [[nodiscard]] const std::string invokeREST(const std::string& url,
+                                             const std::string& httpMethod,
+                                             const std::string& requestBody);
```
rwalworth, line 47: Any specific reason these functions return `const std::string` instead of just `std::string`?

rwalworth (commented):
Did you test this at all? May not be a bad idea to throw a test or two in integration tests to verify functionality.

Also just FYI we generally don't include the PRs in the milestones as the PRs are reachable via their corresponding issues in the milestone. I went ahead and removed the milestone from this.

sdk/main/include/impl/HttpClient.h
rwalworth: There should be header guards as well.

sdk/main/include/impl/HttpClient.h
gsstoykov: You are right, maybe the function user would want to alter the returned string.

sdk/main/src/impl/HttpClient.cc
gsstoykov: Which header do you mean?  #include "impl/HttpClient.h" is in the HttpClient.cc.

sdk/main/include/impl/HttpClient.h
gsstoykov: if you mean the license I added it.

sdk/main/include/impl/HttpClient.h
gsstoykov: Bad comment explanation on my side. We can have several httpClients, but I could not think of a use case in which we need to copy them as they don`t have specific states to be copied and the operations are then useless.

sdk/main/src/impl/HttpClient.cc
rwalworth: Sorry I meant license

sdk/tests/integration/HttpClientIntegrationTests.cc
```diff
@@ -0,0 +1,80 @@
...
+#include <chrono>
+#include <filesystem>
+#include <fstream>
+#include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
+#include <string>
+#include <string_view>
+
+using json = nlohmann::json;
+using namespace std;
```
rwalworth, line 33: Small thing, but if you're `using namespace std` here then you don't need to add the namespace scoping on all `std` variables you use

sdk/tests/integration/HttpClientIntegrationTests.cc
```diff
@@ -0,0 +1,80 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include "BaseIntegrationTest.h"
+#include "impl/HttpClient.h"
+
+#include <chrono>
```
rwalworth, line 24: Is anything in `<chrono>` being used in this file?

config/local_node.json
```diff
@@ -3,7 +3,7 @@
         "0.0.3": "127.0.0.1:50211"
     },
     "mirrorNetwork": [
-        "127.0.0.1:5600"
+        "127.0.0.1:5551"
```
rwalworth, line 6: This will break our other queries that talk directly to the gRPC mirror node endpoint. Since you've only got one test for the `HttpClient` you'd probably be fine to just hardcode this address in the test for now, but if we add more tests down the line we could try to come up with a better way to distinguish between different mirror node endpoints in the config file here. We definitely don't want to change this address here though.

gsstoykov: with the current local node:
Consensus Node Url - 127.0.0.1:50211
Node Id - 0.0.3
Mirror Node Url - http://127.0.0.1:5551

sdk/tests/integration/HttpClientIntegrationTests.cc
gsstoykov: I intended to use timing for the REST request but can be removed! Thanks!

config/local_node.json
rwalworth: Talked offline, but there are several different endpoints for mirror node, `5551` is the REST API while `5600` is the gRPC endpoint. Leaving this as is and will hardcode the REST API URL for now. Will create a new issue down the line to update the config file to have both URLs if need be

rwalworth (approved):
LGTM

PR #558, opened 2023-10-20
Add `StakingWithUpdateExample`

### LINKED ISSUE #557: Add `StakingWithUpdateExample`
Add the `StakingWithUpdateExample`

rwalworth:
**Description**:
This PR adds the `StakingWithUpdateExample`, which shows users how to update their accounts to unstake.

**Related issue(s)**:

Fixes #557 

**Checklist**


sdk/main/src/AccountUpdateTransaction.cc
```diff
@@ -104,6 +104,14 @@ AccountUpdateTransaction& AccountUpdateTransaction::setAccountMemo(std::string_v
   return *this;
 }
 
+//-----
+AccountUpdateTransaction& AccountUpdateTransaction::clearAccountMemo()
+{
+  requireNotFrozen();
+  mAccountMemo = "";
```
gsstoykov, line 111: Since mAccountMemo is an optional, should the clear function not for example set it to std::nullopt. I think it is confusing having an optional string with "" as a value.

gsstoykov (approved):
LGTM.

sdk/main/src/AccountUpdateTransaction.cc
```diff
@@ -128,6 +136,15 @@ AccountUpdateTransaction& AccountUpdateTransaction::setStakedAccountId(const Acc
   return *this;
 }
 
+//-----
+AccountUpdateTransaction& AccountUpdateTransaction::clearStakedAccountId()
+{
+  requireNotFrozen();
+  mStakedAccountId = AccountId(0ULL);
```
deyanzz, line 143: Since `mStakedAccountId` is optional, shouldn't we set it to `std::nullopt` when clearing the `StakedAccountId`?

sdk/main/src/AccountUpdateTransaction.cc
```diff
@@ -138,6 +155,16 @@ AccountUpdateTransaction& AccountUpdateTransaction::setStakedNodeId(const uint64
   return *this;
 }
 
+//-----
+AccountUpdateTransaction& AccountUpdateTransaction::clearStakedNodeId()
+{
+  requireNotFrozen();
+
+  mStakedNodeId = -1LL;
```
deyanzz, line 163: Same as `mStakedAccountId` - since `mStakedNodeId` is optional, shouldn't we set it to `std::nullopt` when clearing the `StakedNodeId`?

deyanzz (approved):

sdk/main/src/AccountUpdateTransaction.cc
rwalworth: I understand the confusion but setting `mAccountMemo` to `std::nullopt` wouldn't update the memo at all which is not desired behavior. Since `AccountUpdateTransaction` can be used to update as many or as little fields as we want, `std::optional` types are used to indicate if a value should be updated or not (i.e. if a `std::optional` field does not have a value, that field should not be updated) (this can be confirmed by looking at the `build()` function). So we want to make sure `mAccountMemo` is initialized with a value so that it gets built into the `AccountUpdateTransaction`'s corresponding protobuf when it gets executed, but we want it to be empty i.e. cleared. If `mAccountMemo` was set to `std::nullopt` here, then no change would be built into the protobuf and the account memo would not actually be cleared. Does that make sense?

sdk/main/src/AccountUpdateTransaction.cc
rwalworth: Same explanation applies here as it does for [`mAccountMemo`](https://github.com/hashgraph/hedera-sdk-cpp/pull/558#discussion_r1368755086).

sdk/main/src/AccountUpdateTransaction.cc
rwalworth: Same explanation applies here as it does for [`mAccountMemo`](https://github.com/hashgraph/hedera-sdk-cpp/pull/558#discussion_r1368755086).

sdk/main/src/AccountUpdateTransaction.cc
gsstoykov: I understand now. Thanks!

sdk/main/src/AccountUpdateTransaction.cc
deyanzz: Thanks for the explanation. I agree this implementation is the correct one.

PR #567, opened 2023-10-25
Fix `AccountId` `evmAddress` field not being displayed correctly in `toSolidityAddress()` 

gsstoykov:
**Description**:
The ```evmAddress``` field was not populated correctly with the an EVM Address when creating
an account with an alias so it had to be fetched from the ```MirrorNode```. ```toSolidityAddress```
function was not returning correct values.

**Related issue(s)**:

**Fixes**  https://github.com/hashgraph/hedera-sdk-cpp/issues/498

**Notes for reviewer**:

**Checklist**


deyanzz (approved):
LGTM

PR #569, opened 2023-10-26
Full implementation of `Executable`

### LINKED ISSUE #566: Full implementation for `Executable`
Implement all the public APIs that are a part of `Executable`, primarily the async operations

rwalworth:
**Description**:
This PR fills out the remaining APIs of the `Executable` base class, primarily the `executeAsync` calls. It also changes all `std::chrono::duration<double>` instances over to `std::chrono::system_clock::duration`, which still provides the desired generality and causes less need for `duration_cast` calls. 

**Related issue(s)**:

Fixes #566 

**Notes for reviewer**:
This adds a good amount of integration tests that may slow down builds on GitHub by a couple of minutes.

**Checklist**

### DIFF
--- a/sdk/main/include/Defaults.h
+++ b/sdk/main/include/Defaults.h
@@ -39,6 +39,10 @@ constexpr auto DEFAULT_MIN_BACKOFF = std::chrono::milliseconds(250);
  */
 constexpr auto DEFAULT_MAX_BACKOFF = std::chrono::seconds(8);
 /**
+ * The default maximum amount of time to spend on one execution attempt
+ */
+constexpr auto DEFAULT_GRPC_DEADLINE = std::chrono::seconds(10);
+/**
  * The default maximum number of times a node is allowed to return a bad gRPC status before it is permanently removed
  * from a network. 0 indicates there's no maximum.
  */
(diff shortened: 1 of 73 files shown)


sdk/main/src/PrivateKey.cc
```diff
@@ -190,6 +190,7 @@ std::vector<std::byte> PrivateKey::signTransaction(WrappedTransaction& transacti
     case TRANSFER_TRANSACTION:
       return signTransaction(*transaction.getTransaction<TransferTransaction>());
     case UNKNOWN_TRANSACTION:
+    default:
```
rwalworth, line 193: This fixes a build warning, unrelated to this PR

sdk/main/include/Defaults.h
```diff
@@ -38,6 +38,10 @@ constexpr auto DEFAULT_MIN_BACKOFF = std::chrono::milliseconds(250);
  * The default maximum duration of time to wait before retrying to submit a previously-failed request.
  */
 constexpr auto DEFAULT_MAX_BACKOFF = std::chrono::seconds(8);
+/**
+ * The default maximum amount of time to spend on one execution attempt
+ */
+constexpr auto DEFAULT_GRPC_DEADLINE = std::chrono::seconds(10);
```
gsstoykov, line 44: Just curious. Is this by specification? Can we pick a better value or have some kind of logic regarding the amount of time spent?

gsstoykov (approved):
LGTM.

sdk/main/include/Defaults.h
rwalworth: Not 100% sure, just stole this from the other SDKs. I'm assuming there was some discussion at some point about this or there was some testing done to determine this being the best value. @SimiHunjan may be able to provide more insight

Issue #423, opened 2023-07-14
OpenSSL keys import compatibility


dikel:
### Problem

Currently when importing ECDSA private keys the DER header prefix is hardcoded but the ECPrivateKey format specified in [RFC-5915 page 2](https://www.rfc-editor.org/rfc/rfc5915) has optional field that will change the prefix if they are present. This makes the SDK not completely compatible with keys generated by OpenSSL. Also there was an issue in Java, JS and Go SDKs (and probably other) and they were exporting the ECDSA keys as PrivateKey format instead of ECPrivateKey and the C++ SDK should be able to import them as well. There are test vectors in the [reference repo](https://github.com/hashgraph/hedera-sdk-reference/issues/93) to ensure the correct implementation.

### Solution

- Make the SDK able to import ECDSA DER/PEM keys generated by OpenSSL
- Add support for DER/PEM keys exported by the old SDK versions for backward compatibility
- Add all test vectors from the reference repo

```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-sdk-cpp/issues/619
```

gsstoykov:
Created an example solution to integrating the keys to the SDK. This ticket should be split due to impact of the changes to the project. Should be split into Integration of ECDSA PrivateKey, ECDSA PublicKey. Also it is best to have a separate issue for reworking the project tests as they would be impacted with big changes as well. **Overall this ticket leads to large changes on the code base and dependencies and should be decomposed.**

gsstoykov:
With the completion of https://github.com/hashgraph/hedera-sdk-cpp/issues/619 this issue can now be closed!

PR #607, opened 2023-11-14
Added EC Private Keys compatibility to improve OpenSSL Key import compatibility.

gsstoykov:
**Description**:

This PR introduces basic logics for working with ASN1 formed data. SDK should support imports from DER/PEM format
and should be fully OpenSSL compatible. This PR demonstrates integrating different types of EC Private Keys. If solution is validated could further push changes for EC Public Keys. Should be relatively easy as a tweak.

**This PR modifies ... in order to support ...**
* PEM/DER Imported EC Private Keys

**Related issue(s)**: #423 

**Checklist**


rwalworth (approved):
Looks like there are some files where maybe the `.clang-format` file didn't format correctly, but other than that LGTM

PR #610, opened 2023-11-14
Implement `.env` file for examples

### LINKED ISSUE #606: Setup examples to run off of a `.env` file input

rwalworth:
**Description**:
This PR adds processing for decoding `.env` files for running examples. This is more straightforward than using command line arguments and more closely follows other SDKs.

**Related issue(s)**:

Fixes #606 

**Checklist**


gsstoykov (commented):
Is there a need to call dotenv::init() in every file? Can we have some static init for this on #include for example?

rwalworth:
> Is there a need to call dotenv::init() in every file? Can we have some static init for this on #include for example?

I don't think I entirely understand what you're suggesting. `dotenv::init()` is the call that looks for the `.env` file and populates the environment variables, so yes it has to be called in every example. We could just throw this in a `#include` but I don't necessarily think that would make the code any more readable or optimized. Unless I'm misunderstanding what you mean?

gsstoykov (approved):
LGTM.

PR #631, opened 2023-11-30
Add `toString()` method and `operator<<` overload to query responses

### LINKED ISSUE #625: QA: Get Token Info
Get the current state of the token from the consensus nodes

rwalworth:
**Description**:
This PR addresses the concerns made in #625, which states that query responses should be able to be printed to an output stream. This involved overloading the `<<` operator for all objects as well as adding missing `toString()` methods to allow for the stringification of these objects, which just puts the objects into a JSON string.

This also adds some other missing APIs for these objects as well, namely `[to|from][Protobuf|Bytes]()`

**Related issue(s)**:

Fixes #625 

**Checklist**

### DIFF
--- a/sdk/examples/AccountAliasExample.cc
+++ b/sdk/examples/AccountAliasExample.cc
@@ -101,8 +101,7 @@ int main(int argc, char** argv)
             << std::endl;
 
   std::cout << "Balance of the created account: "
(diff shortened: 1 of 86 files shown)


gsstoykov (approved):
LGTM.

PR #644, opened 2024-01-10
Add TCK server

### LINKED ISSUE #449: Set-up JSON-RPC Server: C++
In order to interact with the [Hedera TCK](https://github.com/hashgraph/hedera-sdk-tck), the C++ SDK needs a server with which to process JSON-RPC requests and run the corresponding SDK functions, as well as reply back to the TCK with the request status.

rwalworth:
**Description**:
This PR adds the initial processing for the C++ SDK TCK server. It is able to run tests currently in the TCK and produce correct results. As more TCK tests are added, more methods will need to be added and possible arranged into multiple files to keep them neat and organized.

**Related issue(s)**:

Fixes #449 

**Checklist**


tck/src/main.cc
```diff
@@ -0,0 +1,55 @@
...
+using namespace Hedera::TCK;
+
+int main(int argc, char** argv)
+{
+  // Initialize the TCK server with the port number.
+  TckServer tckServer((argc > 1) ? static_cast<int>(Hedera::internal::EntityIdHelper::getNum(argv[1])) : // NOLINT
+                        TckServer::DEFAULT_HTTP_PORT);
+
+  // Add the SDK client functions.
+  tckServer.add("createAccount",
```
gsstoykov, line 34: Just a little question here. Will it be better for the `tckServer` to resolve the `getHandles`? I can see that most of the time the `.add` takes similar names for the first argument which is name and then calls `getHandle` for the same name? And just was wondering shouldn't we define the SDK client functions in the `tckServer` not having to add them every time one by one or is the idea here to have like different suites of tests and that is why you decided to leave it so we can add different SDK functions independently?

gsstoykov (approved):
LGTM.

tck/src/main.cc
rwalworth: I styled a lot of this implementation off of the `json-rpc-cxx` [library](https://github.com/jsonrpcx/json-rpc-cxx) so didn't think too much deeper about the implementation here. Since C++ doesn't support reflection, there needs to be some sort of mapping from the function name string to the actual function. The `getHandle` function also helps differentiate between JSON RPC methods and notifications, which we may or may not use notifications ever but the functionality doesn't hurt to have I think.

As for `SdkClient`, the idea was to have all SDK-facing functionality in its own object (or namespace in this matter, since `SdkClient` is acting like a singleton). But yes, it would be a good idea at some point when more tests are added to possibly make up different test suites, which would also provide more organization to the code.

gsstoykov: Nice we can keep the test suite thing as a future idea.

Issue #462, opened 2023-08-16
HIP-745: Design feature


SimiHunjan:
Design the feature documented in [HIP-745](https://hips.hedera.com/hip/hip-745)

PR #647, opened 2024-01-22
HIP-745 Design Feature

gsstoykov:
**Description**:
Added changes described in https://www.notion.so/limechain/HIP-745-43c4ac8fdded4357b47bac0dbf217638

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-cpp/issues/462

**Notes for reviewer**:
For the most part should be working but not extensively tested and documented so this PR will be updated. PR serves mostly as a draft for other SDKs to see.

**Checklist**


sdk/main/src/Transaction.cc
```diff
@@ -763,9 +794,7 @@ Transaction<SdkRequestType>::Transaction(
 
   mImpl->mTransactionMemo = mImpl->mSourceTransactionBody.memo();
 
-  // This constructor is only used in fromBytes(), which means a Transaction was frozen and serialized using toBytes().
-  // As such, this Transaction should be constructed as frozen.
-  mImpl->mIsFrozen = true;
+  mImpl->mIsFrozen = shouldFreeze;
 }
```
rwalworth, line 798: I'm a bit confused by the added logic in this function. If a `proto::Transaction` doesn't contain signed body bytes, that means the `Transaction` shouldn't be frozen? But will this ever happen? If you use `toBytes` will that ever create an empty `proto::Transaction`? It seems like the number of signatures added to the `proto::Transaction` should be looked at here somewhere, it was my understanding that if a signature has been added to a `proto::Transaction` then it should be frozen, otherwise it shouldn't.

sdk/main/src/Transaction.cc
```diff
@@ -788,6 +817,41 @@ void Transaction<SdkRequestType>::buildAllTransactions() const
...
+         getNodeAccountIds())
+  {
+    // Since transaction proto bodybytes are deprecated we are working around with
+    // the use of signedTransaction proto bodybytes
+    mImpl->mSourceTransactionBody.set_allocated_nodeaccountid(accountId.toProtobuf().release());
+    proto::SignedTransaction signedTransaction;
+    signedTransaction.set_bodybytes(mImpl->mSourceTransactionBody.SerializeAsString());
+    // populating mTransactions which should be further serialized
+    addTransaction(signedTransaction);
+  }
```
rwalworth, line 852: This `for`-loop looks like the same processing that's contained in `addSignedTransactionForEachNode()`, could it be replaced with that?

sdk/main/src/Transaction.cc
```diff
@@ -844,6 +908,12 @@ void Transaction<SdkRequestType>::addTransaction(const proto::Transaction& trans
   signedTx.ParseFromArray(transaction.signedtransactionbytes().data(),
                           static_cast<int>(transaction.signedtransactionbytes().size()));
 
+  // will be true if the transaction came from bytes
+  if (signedTx.ByteSizeLong() == 0)
+  {
+    signedTx.set_bodybytes(transaction.SerializeAsString());
+  }
```
rwalworth, line 915: This doesn't make sense to me. If the `signedTx` doesn't contain any body bytes (which I'm still confused on how that can happen), then you fill it with the bytes of the serialized `transaction`, which contains the serialized `signedTx`?

sdk/main/src/Transaction.cc
```diff
@@ -852,7 +922,11 @@ void Transaction<SdkRequestType>::addTransaction(const proto::Transaction& trans
 template<typename SdkRequestType>
 void Transaction<SdkRequestType>::addTransaction(const proto::SignedTransaction& transaction)
 {
-  mImpl->mTransactions.push_back(proto::Transaction());
+  proto::Transaction tx;
+  tx.ParseFromString(transaction.bodybytes());
```
rwalworth, line 926: Can you create a `proto::Transaction` successfully from a serialized `proto::SignedTransaction`? I would think here it needs to be
```
proto::Transaction tx;
tx.signedtransactionbytes = transaction.SerializeAsString();
```

sdk/main/src/Transaction.cc
```diff
@@ -852,7 +922,11 @@ void Transaction<SdkRequestType>::addTransaction(const proto::Transaction& trans
...
+  // proto::TransactionBody txBody;
+  // txBody.ParseFromArray(transaction.bodybytes().data(), static_cast<int>(transaction.bodybytes().size()));
```
rwalworth, line 928: Remove commented code

sdk/main/src/Transaction.cc
```diff
@@ -949,6 +1023,7 @@ template<typename SdkRequestType>
 proto::Transaction Transaction<SdkRequestType>::getTransactionProtobufObject(unsigned int index) const
 {
   return mImpl->mTransactions.at(index);
+  //** should have this assigned even in not frozen transactions
```
rwalworth, line 1026: What is this comment suggesting?

sdk/examples/AccountAllowanceSerializedExample.cc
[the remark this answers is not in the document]
gsstoykov: I just decided on reusing most of the AccountAllowanceExample as the example is descriptive and I just further added some serialisation/deserialisation. This way we have a reference to the same example without serialisation/deserialisation and this would prove that the flow is similar and the changes don't impact the result of the transaction being executed.

sdk/main/src/Transaction.cc
gsstoykov: Changed the logic for shouldFreeze now it freezes on sigmap empty.

sdk/main/src/Transaction.cc
gsstoykov: It has some differences so I decided not to reuse the function, because that would mean reworking it.

sdk/examples/AccountAllowanceSerializedExample.cc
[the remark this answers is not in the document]
rwalworth: Would showing this functionality possibly be better in the `MultiAppTransferExample`? That example's main purpose is to show how to serialize/deserialize transactions already. I just worry about reusing `AccountAllowanceExample` because it's already a bit of a cumbersome example with a good amount of steps and transactions to show.

PR #682, opened 2024-04-16
feat(android): Add Android builds to CI

### LINKED ISSUE #660: Add Android and iOS builds
It has been requested from the community to add Android and iOS builds for the C++ SDK. These ideally should be added as CMake presets like we already have for `macos-arm64`, `windows-x64`, and `linux-x64`. A way to actually test these builds will also need to be found and added.

rwalworth:
**Description**:
This PR adds Android builds of the C++ SDK to the Github workflows. It will only generate the builds after a PR is merged as to not slow down development. A couple notes:
 - `vcpkg` needed to upgraded as the baseline we were using was too old for Android builds.
 - Switched to using `spdlog` logging library in place of `log4cxx` simply because `vcpkg` currently supports `spdlog` but not `log4cxx` android builds.
 - The Android build is currently done in a docker container, but it could be possibly be done straight on our Github runners as they are `linux-x86_64` if this wanted to be looked into later. I tried this but there was an issue finding `git` that I couldn't figure out quickly.
 - There is an issue testing the Android test executables on an emulator running in an `x86_64` environment; even running an `arm64` docker container in an `x86_64` host environment there were issues. If we would like to test these Android builds in CI, this will need to be figured out. However, I left instructions in the Dockerfile as to how to run the test executables.
 - Some work was done that involved building the examples into Android APKs. This was eventually abandoned, but I left some of the work I had done for that in the Dockerfile as well in case this wants to be done in the future.
 - The examples executables were being installed twice unnecessarily so I removed this.
 - We are now installing `gtest` via `vcpkg` instead of manually through CMake.
 - Some filepaths needed updating.
 - Many files had unnecessary includes that I removed.

**Related issue(s)**:

Fixes #660 

**Checklist**


Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+RUN cmake --build --preset android-arm64-debug
+
+# TODO:
+# The steps below are steps needed to test the Android library executables. Currently, the Android emulator is unable to
+# run in a linux-x86_64 environment because it is built for arm64. If there is interest in testing the Android library
+# in CI, this issue of where to run the Android emulator will need to be figured out (must be in an arm64 environment,
+# arm64 docker container has been tested but doesn't work). The steps here should be the general workflow of how to get
+# the emulator up and running and the test executables ran.
+
+# Download Android SDK
```
gsstoykov, line 65: Maybe these can go into a makefile so we don't leave them just commented out in this docker file.

Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+ENV PATH $ANDROID_NDK_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin:$PATH
+
+# Clone the repository.
+RUN git clone https://github.com/hashgraph/hedera-sdk-cpp.git
+
+# Change the working directory to that of the cloned repository.
+WORKDIR /hedera-sdk-cpp
+
+# Checkout the working branch.
+RUN git checkout origin/$BRANCH_NAME
```
gsstoykov, line 47: Can you combine the below runs into one? Something like:
```
RUN git clone https://github.com/hashgraph/hedera-sdk-cpp.git && \
    cd hedera-sdk-cpp && \
    git checkout origin/$BRANCH_NAME && \
    git submodule update --init && \
    cmake --preset android-arm64-release && \
    cmake --preset android-arm64-debug && \
    cmake --build --preset android-arm64-release && \
    cmake --build --preset android-arm64-debug
```

src/sdk/main/include/Logger.h
```diff
@@ -134,17 +132,17 @@ class Logger
   /**
    * The wrapped log4cxx logger.
```
gsstoykov, line 133: nit: The comment is still referring to `log4cxx`.

Dockerfile
gsstoykov, line 1: How much time does this usually take to run? Curious because I can see a lot of dependencies in there.

gsstoykov (commented):
Overall LGTM but shouldn't we wait and see how the communication with the community member goes before pushing this into `main`? As I saw we still don't have answers from the community member and we would only introduce more configurations and docker pipeline overhead.

Dockerfile
rwalworth: When/if these are implemented, the commands are going to have to be run in the Dockerfile anyways so I don't really see any sense in moving them, unless I'm misunderstanding?

Dockerfile
rwalworth: Typically about ~30 minutes, which is why I only ran the build on merge as to not waste development time

rwalworth:
> Overall LGTM but shouldn't we wait and see how the communication with the community member goes before pushing this into `main`? As I saw we still don't have answers from the community member and we would only introduce more configurations and docker pipeline overhead.

I could go either way. I have the Android builds only being generated on merge so it won't impact development time. Maybe @SimiHunjan can weigh in here.

gsstoykov (approved):
LGTM.

rwalworth:
Not merging for the time being, leaving work on protected branch however

PR #754, opened 2024-08-08
feat(HIP-869): Added DAB Node operations 

gsstoykov:
**Description**:
This PR will introduce the `Node` functionalities defined in https://hips.hedera.com/hip/hip-869.

**Related issue(s)**:
https://github.com/hashgraph/hedera-sdk-cpp/issues/712

Fixes https://github.com/hashgraph/hedera-sdk-cpp/issues/713 https://github.com/hashgraph/hedera-sdk-cpp/issues/714 https://github.com/hashgraph/hedera-sdk-cpp/issues/715 https://github.com/hashgraph/hedera-sdk-cpp/issues/716 https://github.com/hashgraph/hedera-sdk-cpp/issues/717

**Notes for reviewer**:
This PR cannot be integration tested as there is no proper way of testing Node specific transactions at this point.

**Checklist**


src/sdk/main/include/NodeCreateTransaction.h
```diff
@@ -0,0 +1,326 @@
...
+   */
+  NodeCreateTransaction& setAccountId(const AccountId& accountId);
+
+  /**
+   * Set the description for the node.
+   *
+   * @param description A string containing the description.
+   * @return A reference to this NodeCreateTransaction with the newly-set description.
+   */
+  NodeCreateTransaction& setDescription(const std::optional<std::string>& description);
```
rwalworth, line 153: Not sure we'd want to input an `optional` here.. if the user doesn't want to set a description they can just not set one.

src/sdk/main/include/NodeUpdateTransaction.h
```diff
@@ -0,0 +1,343 @@
...
+   */
+  NodeUpdateTransaction& setAccountId(const AccountId& accountId);
+
+  /**
+   * Set the description for the node.
+   *
+   * @param description A string containing the description.
+   * @return A reference to this NodeUpdateTransaction with the newly-set description.
+   */
+  NodeUpdateTransaction& setDescription(const std::optional<std::string>& description);
```
rwalworth, line 162: Same here, can just feed in a `std::string_view`, doesn't need an `optional`

src/sdk/main/src/Endpoint.cc
```diff
@@ -38,12 +39,14 @@ std::unique_ptr<proto::ServiceEndpoint> Endpoint::toProtobuf() const
...
   proto->set_ipaddressv4(internal::Utilities::byteVectorToString(mAddress.toBytes()));
   proto->set_port(static_cast<int32_t>(mPort));
+  proto->set_domain_name(mDomainName);
   return proto;
 }
 
 //-----
 std::string Endpoint::toString() const
 {
+
```
rwalworth, line 49: remove this line

src/sdk/main/src/NodeCreateTransaction.cc
```diff
@@ -0,0 +1,195 @@
...
+  if (!transactionBody.has_nodecreate())
+  {
+    throw std::invalid_argument("Transaction body doesn't contain NodeCreate data");
+  }
+
+  const aproto::NodeCreateTransactionBody& body = transactionBody.nodecreate();
+
+  mAccountId = AccountId::fromProtobuf(body.account_id());
+
+  mDescription = body.description();
```
rwalworth, line 139: Should check for `description` validity here (`if (body.has_description()) {...}`)

src/sdk/main/src/NodeCreateTransaction.cc
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  body->set_gossip_ca_certificate(internal::Utilities::byteVectorToString(mGossipCaCertificate));
+
+  if (mGrpcCertificateHash.has_value())
+  {
+    body->set_grpc_certificate_hash(internal::Utilities::byteVectorToString(mGrpcCertificateHash.value()));
+  }
+
+  body->set_allocated_admin_key(mAdminKey->toProtobufKey().release());
```
rwalworth, line 190: Should check to make sure `mAdminKey` is not `nullptr` before filling out

src/sdk/main/src/Transaction.cc
```diff
@@ -355,8 +364,9 @@ std::vector<std::byte> Transaction<SdkRequestType>::toBytes() const
 template<typename SdkRequestType>
 SdkRequestType& Transaction<SdkRequestType>::sign(const std::shared_ptr<PrivateKey>& key)
 {
-  return signInternal(
-    key->getPublicKey(), [key](const std::vector<std::byte>& vec) { return key->sign(vec); }, key);
+  // clang-format off
+  return signInternal(key->getPublicKey(), [key](const std::vector<std::byte>& vec) { return key->sign(vec); }, key);
+  // clang-format on
```
rwalworth, line 369: Is this intentional?

src/sdk/tests/unit/NodeCreateTransactionUnitTests.cc
```diff
@@ -0,0 +1,185 @@
...
+  EXPECT_EQ(nodeCreateTransaction.getGrpcCertificateHash(), bytes);
+  EXPECT_EQ(nodeCreateTransaction.getAdminKey()->toProtobufKey()->SerializeAsString(),
+            key->toProtobufKey()->SerializeAsString());
+}
+
+//-----
+TEST_F(NodeCreateTransactionUnitTests, SetAndGetAccountId)
+{
+  // Given
+  Hedera::AccountId accountId(1234);
```
rwalworth, line 98:
```suggestion
AccountId accountId(1234);
```
nit: already `using namespace Hedera;`

src/sdk/tests/unit/NodeUpdateTransactionUnitTests.cc
```diff
@@ -0,0 +1,187 @@
...
+  EXPECT_EQ(nodeUpdateTransaction.getGrpcCertificateHash(), bytes);
+  EXPECT_EQ(nodeUpdateTransaction.getAdminKey()->toProtobufKey()->SerializeAsString(),
+            key->toProtobufKey()->SerializeAsString());
+}
+
+//-----
+TEST_F(NodeUpdateTransactionUnitTests, SetAndGetAccountId)
+{
+  // Given
+  Hedera::AccountId accountId(1234);
```
rwalworth, line 100:
```suggestion
AccountId accountId(1234);
```
same here

rwalworth (commented):
Looks good overall, just a couple of things to look at

src/sdk/main/src/Transaction.cc
gsstoykov: Yes, this line constantly causes issues for me as I'm using the same clang-format version configured in the CI but the file isn't passing the clang-format check. I always have to manually edit this so it passes... Was thinking of why just not save the effort. If other issues with formatting appear I would investigate and try reworking my local setup. But I guess this doesn't harm the codebase that much.

src/sdk/main/src/NodeCreateTransaction.cc
gsstoykov: Sorry, missed that :)

src/sdk/main/src/NodeCreateTransaction.cc
gsstoykov: There is no `body.has_description()` sadly. Otherwise would have checked.

rwalworth: Hmm maybe `mDescription` shouldn't be `optional` then? It doesn't really make sense to have an `optional` object if there's no way to check in the protobuf if it's valid or not

gsstoykov:
```
  /**
   * A short description of the node.
   * This value, if set, MUST NOT exceed 100 bytes when encoded as UTF-8.
   * This field is OPTIONAL.
   */
```
   IDK really lets leave it like that for now

rwalworth (approved):
LGTM

PR #790, opened 2024-10-09
feat(tokenDeleteTransaction): Implement JSON-RPC endpoint for `TokenDeleteTransaction`

### LINKED ISSUE #789: `TCK-C++`: Implement JSON-RPC methods for `TokenDeleteTransaction`
Implement the `deleteToken` endpoint and test the C++ SDK against the tests in the TCK for `TokenDeleteTransaction`.

rwalworth:
**Description**:
This PR implements the JSON-RPC for `TokenDeleteTransaction` (`deleteToken`).

**Related issue(s)**:

This contains work in #793, so that should be reviewed first.

Fixes #789 

**Checklist**


src/tck/src/SdkClient.cc
```diff
@@ -129,6 +136,176 @@ nlohmann::json SdkClient::createAccount(const std::optional<std::string>& key,
...
+                                      const std::optional<int64_t>& autoRenewPeriod,
+                                      const std::optional<std::string>& memo,
+                                      const std::optional<std::string>& tokenType,
+                                      const std::optional<std::string>& supplyType,
+                                      const std::optional<int64_t>& maxSupply,
+                                      const std::optional<std::string>& feeScheduleKey,
+                                      const std::optional<std::vector<std::shared_ptr<CustomFee>>>& customFees,
+                                      const std::optional<std::string>& pauseKey,
+                                      const std::optional<std::string>& metadata,
+                                      const std::optional<std::string>& metadataKey,
```
gsstoykov, line 162: Was thinking about the parameters here and for the other `SdkClient` functions.. Won't it look better to have them defined in something like `CommonTransactionParams` but for example `CreateTokenTransactionParams`. This way we won't have to write those in both the `.h` and `.cc` or potentially somewhere else as arguments. Also this way we would only pass in one reference to the `CreateTokenTransactionParams`. WDYT?

src/tck/src/main.cc
```diff
@@ -44,9 +44,36 @@ int main(int argc, char** argv)
...
+                  "autoRenewPeriod",
+                  "memo",
+                  "tokenType",
+                  "supplyType",
+                  "maxSupply",
+                  "feeScheduleKey",
+                  "customFees",
+                  "pauseKey",
+                  "metadata",
+                  "metadataKey",
```
gsstoykov, line 71: These here could potentially be reduced as well. Related to the other comment.

src/sdk/main/src/CustomRoyaltyFee.cc
```diff
@@ -91,11 +91,6 @@ CustomRoyaltyFee& CustomRoyaltyFee::setNumerator(const int64_t& numerator)
 //-----
 CustomRoyaltyFee& CustomRoyaltyFee::setDenominator(const int64_t& denominator)
 {
-  if (denominator == 0LL)
```
gsstoykov, line 94: Are these removed for a reason? Guessing it is checked service side?

src/tck/src/SdkClient.cc
rwalworth: Yesss I actually had this same thought recently as well, I think it would help clean up files and make it less error-prone. I was planning on fixing this for all of our functions, as well as dividing up the `SdkClient` into more manageable files as well.

src/sdk/main/src/CustomRoyaltyFee.cc
rwalworth: Yes exactly

gsstoykov (approved):
LGTM

gsstoykov (approved):

PR #795, opened 2024-10-11
fix: get rid of `curl` and replace its processing with `cpp-httplib`

### LINKED ISSUE #794: Replace `curl` with `cpp-httplib`
`curl` doesn't appear to have the licensing we properly need to move to Hiero, but `cpp-httplib` does. So we should replace our usages of the `curl` library with `cpp-httplib`, which is already used by the TCK server.

rwalworth:
**Description**:
In order to move the `hedera-sdk-cpp` to Hiero, we must be sure to use libraries and dependencies only with certain licenses. The license offered with `curl` did not meet the proper requirements, so we needed to get rid of it. Fortunately, the `cpp-httplib` which has been utilized by the C++ SDK TCK server could do the same job. This PR removes all traces of `curl` from the repo, and replaces its processing with `cpp-httplib`.

**Related issue(s)**:

Fixes #794 

**Checklist**


src/sdk/main/src/impl/HttpClient.cc
```diff
@@ -18,101 +18,77 @@
...
-// beforehand
-std::string HttpClient::invokeREST(const std::string& url,
-                                   const std::string& httpMethod,
-                                   const std::string& requestBody)
-{
-  std::unique_ptr<CURL, decltype(&curl_easy_cleanup)> curl(curl_easy_init(), curl_easy_cleanup);
+  httplib::Result res;
 
-  if (!curl)
+  // Perform the request based on the HTTP method
```
gsstoykov, line 51: nit: Maybe would look better if we have:
```
const std::unordered_set<std::string_view> supportedMethods = { "GET", "POST" };
```
Check against the set and throw if not in the set.

gsstoykov (commented):
LGTM, Overall just one nit.

src/sdk/main/src/impl/HttpClient.cc
```diff
@@ -18,101 +18,77 @@
...
-  curl_easy_setopt(curl.get(), CURLOPT_WRITEDATA, &response);
-
-  curl_easy_setopt(curl.get(), CURLOPT_CUSTOMREQUEST, httpMethod.c_str());
-
-  if (httpMethod == "POST")
+  else if (method == "POST")
+  {
+    res = client.Post(path, body.data(), body.size(), "application/json");
+  }
+  else
```
gsstoykov, line 60: nit: Maybe would look better if we have:
```
const std::unordered_set<std::string_view> supportedMethods = { "GET", "POST" };
```
Check against the set and throw if not in the set.

rwalworth: I'm not sure if this would be better, unless I'm misunderstanding. We only have two options of what `method` could be: `"GET"` or `"POST"` and `client` needs to do something different based on each one. I can't think of way using a set that would make this more efficient or cleaner/more readable. Maybe we could make the `method` an `enum`? But even with a set, we'd still have to check `if (method == "GET") { ... }` and `else if (method == "POST") { ... }`. Again, unless I'm misunderstanding something

gsstoykov: Come to think of it it really would be more readable like you said.

gsstoykov:
```
if (supportedMethods.find(method) == supportedMethods.end()) {
  throw std::invalid_argument("Unsupported HTTP method: " + std::string(method));
}
```
Was thinking something like this

gsstoykov (approved):

PR #868, opened 2025-01-03
feat: POC refactor for TokenRejectFlow

gsstoykov:
**Description**:
Introduces an example improvement to `TokenRejectFlow` class. There is no need to create overcomplicated class hierarchies as Flows are not Transactions themselves so that approach would be hurtful to our design. Instead I decided that we can just hold transactions themselves as members of the flows and this way we won't have to reimplement existing selectors/mutators and etc. every time we have to create a flow. Also this way we can treat every transaction individually meaning that we can freely set execution and transaction parameters.

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-cpp/issues/822

**Notes for reviewer**:

**Checklist**


src/sdk/main/src/TokenRejectFlow.cc
```diff
@@ -20,125 +20,27 @@ TransactionResponse TokenRejectFlow::execute(const Client& client)
...
-  }
-  else if (mPublicKey && mSigner.has_value())
-  {
-    tokenRejectTransaction.signWith(mPublicKey, mSigner.value());
-  }
-
   // Submit the TokenRejectTransaction.
-  TransactionResponse txResponse = tokenRejectTransaction.execute(client, timeout);
+  TransactionResponse txResponse =
+    tokenRejectTransaction.freezeWith(&client).sign(mReceiverPrivateKey).execute(client, timeout);
```
rwalworth, line 25: Will there be any problems here if `mReceiverPrivateKey` is `null`/not set?

gsstoykov: Added null check on transaction `.sign()`.

rwalworth (approved):
LGTM

PR #878, opened 2025-01-14
Add a query to estimate the gas of a contract call

gsstoykov:
**Description**:

A new `MirrorNodeContractQuery` class has been introduced to support EVM gas estimation and simulation. This query class provides two distinct types of queries:

1. **`MirrorNodeContractCallQuery`** – Executes a contract call and returns the result, simulating the execution of the contract.
2. **`MirrorNodeContractEstimateGasQuery`** – Estimates the gas usage for a contract call, allowing developers to understand the cost of executing a transaction.

The inputs for these queries match those of `ContractExecuteTransaction`, making it easy for developers to calculate the cost of a follow-up transaction.

The `MirrorNodeContractQuery` issues an HTTP request to the `api/v1/contracts/call` endpoint of the Mirror Node Web3 module, which supports both gas estimation and transient simulation of state-changing transactions. This enables accurate cost predictions and helps developers optimise their contract interactions before executing them on the network.

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-cpp/issues/841

**Notes for reviewer**:
Design was agreed upon from discussions for Java/GO but could be reworked in the future as it is not fully extendable. The C++ SDK implements the design so it aligns with the other SDKs.

**Checklist**


src/sdk/examples/MirrorNodeContractQueryExample.cc
```diff
@@ -0,0 +1,102 @@
...
+    .setSender(operatorAccountId)
+    .setGasLimit(30000)
+    .setGasPrice(1234)
+    .setFunction("greet", opt);
+
+  uint64_t estimatedGas = std::stoi(query.execute(client), nullptr, 16);
+  std::cout << "Estimated gas was " << estimatedGas << std::endl;
+
+  // Do the query against the consensus node using the estimated gas
+  ContractFunctionResult contractFunctionResult = ContractCallQuery()
```
rwalworth, line 90: This doesn't look used after initialization

src/sdk/main/include/impl/HttpClient.h
```diff
@@ -2,6 +2,10 @@
 #ifndef HIERO_SDK_CPP_IMPL_HTTP_CLIENT_H_
 #define HIERO_SDK_CPP_IMPL_HTTP_CLIENT_H_
 
+#ifndef CPPHTTPLIB_OPENSSL_SUPPORT
+#define CPPHTTPLIB_OPENSSL_SUPPORT
+#endif
```
rwalworth, line 7: Just curious what's this needed for?

src/sdk/main/include/impl/MirrorNodeContractQuery.h
```diff
@@ -0,0 +1,264 @@
...
+ * MirrorNodeContractQuery returns a result from EVM execution such as cost-free execution of read-only smart
+ * contract queries, gas estimation, and transient simulation of read-write operations.
+ */
+class MirrorNodeContractQuery
+{
+public:
+  /**
+   * Default constructor.
+   */
+  MirrorNodeContractQuery() = default;
```
rwalworth, line 32: You don't need to declare a `default` constructor unless you are providing another

src/sdk/main/include/impl/MirrorNodeContractQuery.h
```diff
@@ -0,0 +1,264 @@
...
+   * @return Reference to the updated object.
+   */
+  MirrorNodeContractQuery& setSenderEvmAddress(const std::string& address);
+
+  /**
+   * Gets the call data for the Mirror Node call.
+   *
+   * @return The call data as a vector of bytes.
+   */
+  const std::vector<std::byte>& getCallData() const { return mCallData; }
```
rwalworth, line 101: I wouldn't return `const` references to internal members. This unnecessarily limits what the user can do with that data.

src/sdk/main/src/impl/MirrorNodeContractCallQuery.cc
```diff
@@ -0,0 +1,38 @@
...
+{
+
+std::string MirrorNodeContractCallQuery::execute(const Client& client)
+{
+  if (!getContractEvmAddress().has_value())
+  {
+    populateContractEvmAddress(client);
+  }
+
+  json contractCallResult =
```
rwalworth, line 18:
```suggestion
  const json contractCallResult =
```

src/sdk/main/src/impl/MirrorNodeContractEstimateGasQuery.cc
```diff
@@ -0,0 +1,40 @@
...
+std::string MirrorNodeContractEstimateGasQuery::execute(const Client& client)
+{
+  setEstimate(true);
+
+  if (!getContractEvmAddress().has_value())
+  {
+    populateContractEvmAddress(client);
+  }
+
+  json contractCallResult =
```
rwalworth, line 20:
```suggestion
  const json contractCallResult =
```

src/sdk/main/src/impl/MirrorNodeContractQuery.cc
```diff
@@ -0,0 +1,144 @@
...
+MirrorNodeContractQuery& MirrorNodeContractQuery::setEstimate(bool estimate)
+{
+  mEstimate = estimate;
+  return *this;
+}
+
+//-----
+void MirrorNodeContractQuery::populateContractEvmAddress(const Client& client)
+{
+  json contractInfo = internal::MirrorNodeGateway::MirrorNodeQuery(client.getClientMirrorNetwork()->getNetwork()[0],
```
rwalworth, line 84:
```suggestion
  const json contractInfo = internal::MirrorNodeGateway::MirrorNodeQuery(client.getClientMirrorNetwork()->getNetwork()[0],
```

src/sdk/main/include/impl/MirrorNodeContractQuery.h
gsstoykov: True, forgot to remove as it was autogen.

src/sdk/examples/MirrorNodeContractQueryExample.cc
gsstoykov: Yes, had to print the result so we can see the simulation result is similar to this call result.

gsstoykov: Reworked! closing

src/sdk/main/include/impl/MirrorNodeContractQuery.h
```diff
@@ -98,7 +91,7 @@ class MirrorNodeContractQuery
    *
    * @return The call data as a vector of bytes.
    */
-  const std::vector<std::byte>& getCallData() const { return mCallData; }
+  std::vector<std::byte> getCallData() const { return mCallData; }
```
rwalworth, line 94: All return types should be like this

gsstoykov: Updated!

rwalworth (approved):
LGTM!

PR #885, opened 2025-01-27
feat(accountApproveAllowanceTransaction): Implement JSON-RPC method endpoint for `AccountApproveAllowanceTransaction`

### LINKED ISSUE #863: `TCK-C++`: Implement JSON-RPC method endpoint for `AccountApproveAllowanceTransaction`
Implement the `approveAllowance` endpoint and test the C++ SDK against the tests in the TCK for `AccountApproveAllowanceTransaction`.

rwalworth:
**Description**:
This PR implements the JSON-RPC for `AccountApproveAllowanceTransaction` (`approveAllowance`).

**Related issue(s)**:

Fixes #863 

**Checklist**


src/sdk/main/include/TokenAllowance.h
```diff
@@ -96,7 +96,7 @@ class TokenAllowance
   /**
    * The amount of tokens that are being allowed to be spent.
    */
-  uint64_t mAmount = 0ULL;
+  int64_t mAmount = 0LL;
```
gsstoykov, line 99: Can a negative amount be allowed to be spent? I see it was unsigned before

gsstoykov (commented):
Was just wondering is there an idea for mock unit testing the SDK TCK servers?

src/sdk/main/include/TokenAllowance.h
rwalworth: It should be allowed to be sent because the protobuf has the amount as an `int64`. Currently the consensus node doesn't allow for negative allowances, so a transaction with a negative amount would fail. However, I think it's important to match the protobuf types as closely as we can. I know there are a couple other types in the SDK that are unsigned when the protobuf should be signed, but yeah I think it makes sense to match the protobuf.

rwalworth:
> Was just wondering is there an idea for mock unit testing the SDK TCK servers?

Haven't thought about it but makes sense. Will make an issue for this

gsstoykov (approved):
LGTM.

PR #895, opened 2025-02-13
feat(HIP-991): Permissionless revenue - Generating Topic Ids for Topic Operators

gsstoykov:
**Description**:

This HIP introduces new transactions and updates to existing transactions in the consensus node software. SDKs will need to support these new transactions and updates to existing ones so that developers can take advantage of the new features.

- [x] Implementation
- [x] Unit tests
- [x] E2E tests
- [x] Example

Also introduced `setTestClientOperator`, `setDefaultTestClientOperator` functions to `BaseIntegration`.

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-cpp/issues/750

**Notes for reviewer**:

* Unit tests had to be reworked to include `transaction.pb.h` as `transaction_body.pb.h` was deleted.
* `AccountStakersQuery` has been dropped from consensus support.

**Checklist**


src/sdk/examples/RevenueGeneratingTopicsExample.cc
```diff
@@ -0,0 +1,224 @@
...
+   */
+  client.setOperator(operatorAccountId, operatorPrivateKey);
+
+  accountBalanceAfter = AccountBalanceQuery().setAccountId(alice).execute(client);
+  feeCollectorBalanceAfter = AccountBalanceQuery().setAccountId(operatorAccountId).execute(client);
+
+  std::cout << "Alice account Hbar balance before: " << accountBalanceBefore.toString() << std::endl;
+  std::cout << "Alice account Hbar balance after: " << accountBalanceAfter.toString() << std::endl;
+  std::cout << "Fee collector account Hbar balance before: " << feeCollectorBalanceBefore.toString() << std::endl;
+  std::cout << "Fee collector account Hbar balance after: " << feeCollectorBalanceAfter.toString() << std::endl;
```
rwalworth, line 162: Should read the token balance, not hbar

src/sdk/main/include/Transaction.h
```diff
@@ -233,6 +234,32 @@ class Transaction
...
+   */
+  SdkRequestType& addCustomFeeLimit(const CustomFeeLimit& customFeeLimit);
+
+  /**
+   * Clear all maximum custom fees that the user is willing to pay for the transaction.
+   *
+   * @return A reference to this derived Transaction object with cleared custom fee limits.
+   * @throws IllegalStateException If this Transaction is frozen.
+   */
+  SdkRequestType& clearCustomFeeLimits();
```
rwalworth, line 261: I believe for now, since `TopicMessageSubmitTransaction` is the only transaction that uses this value, we were going to implement these functions there.

src/sdk/main/src/CustomFeeLimit.cc
```diff
@@ -0,0 +1,106 @@
...
+    if (i > 0)
+    {
+      result += ", ";
+    }
+    result += mCustomFees[i].toString();
+  }
+
+  result += "]}";
+
+  return result;
```
rwalworth, line 104: Other objects use `nlohmann::json` to simplify this processing. I think that should be used here as well, in addition to keeping it uniform

src/sdk/main/include/Transaction.h
gsstoykov: The `custom_fee_limits` defined in the protobufs belongs to `transaction.proto` https://github.com/hashgraph/hedera-protobufs/blob/c0ca3524f2e80e5d5e545e36fcb5d23d64c31fb5/services/transaction.proto and has no corresponding entry in `consensus_submit_message.proto`. I'm guessing this is decided for further extendability as some other transaction might benefit from the functionality. If I work around the implementation to be explicitly present in `TopicSubmitMessage` maybe this would have to be reworked at some future point.

src/sdk/examples/RevenueGeneratingTopicsExample.cc
gsstoykov: I got the message wrong here. Otherwise I modified the `toString()` method to display both `hbar` and tokens if any. Does that make sense to you?

rwalworth: Yeah I just meant the message should read `Fee collector account token balance before/after:`

src/sdk/main/include/Transaction.h
gsstoykov: Changed!

rwalworth (approved):
LGTM

PR #906, opened 2025-02-24
feat: Remove zero shard and realm assumption

gsstoykov:
**Description**:

* `populateAccountEvm` now sets the full `accountId` for the mirror node query.
*  The flagged `zeroAddress` check is now removed as the assumption for 12 zero bytes would be wrong if there are populated shard and realm values.

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-cpp/issues/898

**Notes for reviewer**:

**Checklist**


src/sdk/main/include/impl/EntityIdHelper.h
```diff
@@ -109,14 +109,6 @@ template<typename T = uint64_t>
  */
 [[nodiscard]] std::string checksum(std::string_view address, const LedgerId& ledgerId);
 
-/**
- * Determine if the input byte array could represent a long-zero address.
- *
- * @param address The byte array to determine if it could be a long-zero address.
- * @return \c TRUE if the input byte array could represent a long-zero address, otherwise \c FALSE.
- */
-[[nodiscard]] bool isLongZeroAddress(const std::vector<std::byte>& address);
```
rwalworth, line 118: What implications does this have of just removing it? Can we no longer tell the difference between a long-zero address vs. an EVM address?

rwalworth (approved):

PR #910, opened 2025-03-04
feat(HIP-1021): Improve assignment of auto renew account id for topics

gsstoykov:
**Description**:

Introduces automatic AutoRenewAccountId assignment logic as described in https://github.com/hashgraph/hedera-improvement-proposal/pull/1021

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-cpp/issues/900

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/sdk/main/include/TokenCreateTransaction.h
+++ b/src/sdk/main/include/TokenCreateTransaction.h
@@ -174,14 +174,6 @@ public:
   TokenCreateTransaction& setAutoRenewAccountId(const AccountId& accountId);
 
   /**
-   * Assignment logic of the autoRenewAccount for a transaction.
-   *
-   * @param body The body of the transaction that requires custom AutoRenewAccount assignment logic.
-   * @param accountId The auto renew account to be assigned.
-   */
-  void assignAutoRenewAccount(proto::TransactionBody& body, const AccountId& accountId) const override;
-
-  /**
    * Set the desired auto-renew period for the new token.
    *
    * If expirationTime is set - autoRenewPeriod will be effectively ignored and
--- a/src/sdk/main/include/TokenCreateTransaction.h
+++ b/src/sdk/main/include/TokenCreateTransaction.h
@@ -176,7 +176,8 @@ public:
   /**
    * Assignment logic of the autoRenewAccount for a transaction.
    *
-   * @param account The auto renew account to be assigned.
+   * @param body The body of the transaction that requires custom AutoRenewAccount assignment logic.
+   * @param accountId The auto renew account to be assigned.
    */
   void assignAutoRenewAccount(proto::TransactionBody& body, const AccountId& accountId) const override;
 
--- a/src/sdk/main/include/TokenCreateTransaction.h
+++ b/src/sdk/main/include/TokenCreateTransaction.h
@@ -154,6 +154,12 @@ public:
   /**
    * Set the desired expiration time for the new token.
    *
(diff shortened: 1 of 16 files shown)


src/sdk/main/include/TokenCreateTransaction.h
```diff
@@ -167,9 +173,20 @@ class TokenCreateTransaction : public Transaction<TokenCreateTransaction>
    */
   TokenCreateTransaction& setAutoRenewAccountId(const AccountId& accountId);
 
+  /**
+   * Assignment logic of the autoRenewAccount for a transaction.
+   *
+   * @param body The body of the transaction that requires custom AutoRenewAccount assignment logic.
+   * @param accountId The auto renew account to be assigned.
+   */
+  void assignAutoRenewAccount(proto::TransactionBody& body, const AccountId& accountId) const override;
```
rwalworth, line 182: Should this be declared as `protected` instead of `public`? Since it's only used internally by `Transaction`?

src/sdk/main/src/Transaction.cc
```diff
@@ -947,6 +947,12 @@ void Transaction<SdkRequestType>::updateSourceTransactionBody(const Client* clie
 
   // Add derived Transaction fields to mSourceTransactionBody.
   addToBody(mImpl->mSourceTransactionBody);
+
+  // Assign auto-renew account if the transaction is auto-renew assignable
+  if (client && client->getOperatorAccountId().has_value() && autoRenewAssignable(mImpl->mSourceTransactionBody))
```
rwalworth, line 952: I'm not sure if this `if` check is correct, what if the user assigns the transaction ID manually to an account that isn't the operator? I think it should use the `mImpl->mTransactionId.mAccountId`

src/sdk/main/include/Transaction.h
```diff
@@ -450,6 +450,22 @@ class Transaction
    */
   virtual void validateChecksums(const Client& client) const = 0;
 
+  /**
+   * Assignment logic of the autoRenewAccount for a transaction.
+   *
+   * @param body The body of the transaction that requires custom AutoRenewAccount assignment logic.
+   * @param accountId The auto renew account to be assigned.
+   */
+  virtual void assignAutoRenewAccount(proto::TransactionBody& body, const AccountId& accountId) const;
```
rwalworth, line 459: Do we need a new `virtual` function? Could the required processing for these just be added to the `addToBody` call for the transactions that need this, using the transaction ID already set in the `proto::TransactionBody`?

src/sdk/main/src/TokenCreateTransaction.cc
```diff
@@ -252,6 +252,12 @@ void TokenCreateTransaction::validateChecksums(const Client& client) const
 void TokenCreateTransaction::addToBody(proto::TransactionBody& body) const
 {
   body.set_allocated_tokencreation(build());
+
+  if (body.has_transactionid() && !body.tokencreation().has_autorenewaccount())
+  {
+    std::unique_ptr<proto::AccountID> accountId = std::make_unique<proto::AccountID>(body.transactionid().accountid());
+    body.mutable_tokencreation()->set_allocated_autorenewaccount(accountId.release());
```
rwalworth, line 259: Might want to double check, but I think I remember hearing something about how this only should happen if the admin key isn't set. If that's true we should add a check for that in the `if`. Same with topic create

rwalworth (approved):
One thing to double check that I forgot to mention, otherwise LGTM!

PR #1081, opened 2026-01-23
ci: add markdown link checker workflow

Egbaiyelo:
**Description**:

Adds a ci workflow for checking broken markdown links

**Related issue(s)**:
Issue #1060

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/flow-pull-request-checks.yaml
+++ b/.github/workflows/flow-pull-request-checks.yaml
@@ -38,5 +38,7 @@ jobs:
     - name: Check Markdown links
       uses: gaurav-nelson/github-action-markdown-link-check@5c5dfc0ac2e225883c0e5f03a85311ec2830d368 # v1
       with:
+        check-modified-files-only: 'yes'
+        config-file: '.mlc_config.json'
         use-quiet-mode: 'yes'
         base-branch: 'main'
--- a/.mlc_config.json
+++ b/.mlc_config.json
@@ -1,0 +1,5 @@
+{
+  "retryOn429": true,
+  "retryCount": 3,
+  "timeout": "20s"
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,7 +102,7 @@ emoji ➡️ followed by a description in plain English for each over-arching ne
 #### Examples
 
 This section should list all of the newly added examples. Each of them should represent a link to the implementation (
-for example: [TransferTokensExample.cpp](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/src/sdk/examples/TransferTokensExample.cpp). )It's a good idea also to add prefix with ➡️ before
+for example: [TransferTokensExample.cpp](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/src/sdk/examples/TransferTokensExample.cpp) ). It's a good idea also to add prefix with ➡️ before
 each line.
 
 #### Notes


rwalworth (dismissed):
Hey @Egbaiyelo, great work on getting through all of this! I know fixing commits, rebasing, dealing with merge conflicts, etc. isn't the most exciting work by any means, but you have got through it all and figured it out, I very much admire the drive! I'm going to run the workflows now and we can look to get this merged.

rwalworth:
Alright so looks like the workflow is running correctly, that's awesome to see! However, it is failing, so let's walk through what's happening.

### Issue 1: Missing config file

The workflow still references a config file that no longer exists:

```
Cannot find mlc_config.json
```

Looking at runner logs, the action is looking for `config-file: mlc_config.json`, but the file was removed.

### Issue 2: Rate limiting on unrelated file

The actual failure:

```
FILE: ./docs/training/blogs.md
[✖] https://github.com/hiero-ledger/hiero-website/blob/main/README.md → Status: 429
ERROR: 1 dead links found!
```

The workflow failed because GitHub returned a **429 (Too Many Requests)** when checking a link in `docs/training/blogs.md`, which is a file this PR didn't modify.

This is happening because `check-modified-files-only` was removed, so the checker now scans ALL markdown files in the repo. When it does that, it hits GitHub's rate limits, and unrelated files cause the PR to fail.

### Recommended Fix

Restore the approach from the reference workflow in the issue:

1. **Re-add `check-modified-files-only: 'yes'`** to the workflow. This ensures only files changed in the PR are checked, avoiding rate limits and preventing unrelated failures.

2. **Keep a `.mlc_config.json`** to handle transient failures:

```json
{
  "retryOn429": true,
  "retryCount": 3,
  "timeout": "20s"
}
```

This matches what the original issue specified and avoids the rate limiting problem.

### Minor: Typo in CONTRIBUTING.md

There's still a stray `)` after the period on line 105:

```markdown
[TransferTokensExample.cpp](...). )It's a good idea
```

Should be:

```markdown
[TransferTokensExample.cpp](...). It's a good idea
```

---

Let me know if you have questions!

Egbaiyelo:
I did delete the mlc config file and the reference to it and I thought it was suggestive in the issue, not mandatory my bad, same with the `check-modified-files-only` - I used my discretion here as I realized that links could break and we'd want them to be fixed but I realize the committer shouldn't have to deal with that. 

The trailing ) was to close the one around example like in other places
image
I will realign it so it looks better actually, If you want it removed I will
image
My preview is squished but its just an `English bracket` around an example I believe

yes, CI is really daunting

.github/workflows/flow-pull-request-checks.yaml
```diff
@@ -23,3 +23,22 @@ jobs:
...
+    - name: Harden the runner (Audit all outbound calls)
+      uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
+      with:
+        egress-policy: audit
+
+    - name: Checkout Code
+      uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+
+    - name: Check Markdown links
+      uses: gaurav-nelson/github-action-markdown-link-check@5c5dfc0ac2e225883c0e5f03a85311ec2830d368 # v1
```
rbarker-dev: I have requested a step-security maintained version of this action. For now this is fine.

Issue #1104, opened 2026-02-01
[Good First Issue]: HbarUnit.h uses uint64_t but does not include <cstdint>, which causes build failures on stricter compilers (GCC / Linux / WSL).


tech0priyanshu:
### 🆕🐥 First-Time Friendly

This issue is especially welcoming for people who are new to contributing to the **Hiero C++ SDK**.

We know that opening your first pull request can feel like a big step.  
Issues labeled **Good First Issue** are designed to make that experience easier, clearer, and more comfortable.

No prior knowledge of Hiero, Hedera, or distributed ledger technology is required.  
Just a basic familiarity with C++ and Git is more than enough to get started.

### 👾 Description of the Issue

Building the project on Linux (GCC, C++17) fails due to a missing standard include.

The file `src/sdk/main/include/HbarUnit.h` uses the type `uint64_t` but does not include the required header `<cstdint>`.

On stricter toolchains (such as GCC on Linux), this results in a compilation error because `uint64_t` is not defined.

This compiles successfully on some platforms (like macOS/Clang) due to indirect includes from other headers, but this behavior is not guaranteed by the C++ standard.

**File to Change:**
- `src/sdk/main/include/HbarUnit.h`

### 💡 Proposed Solution

Add the missing `<cstdint>` include to ensure the `uint64_t` type is properly defined on all platforms.

This is a single-line addition — no other changes are needed.

### 👩‍💻 Implementation Steps (End-to-End)

Follow these steps to complete this issue:

1. **Open the file:**  
   `src/sdk/main/include/HbarUnit.h`

2. **Locate the existing includes** (near the top of the file, around lines 24–26):
   ```cpp
   #include <string>
   #include <string_view>
   ```

3. **Add the missing include:**  
   Insert `#include <cstdint>` with the other standard library includes:
   ```cpp
   #include <cstdint>
   #include <string>
   #include <string_view>
   ```

4. **Save the file.**

5. **Verify the change:**  
   Build the project to confirm no compilation errors occur:
   ```bash
   cmake --preset linux-x64-release
   cmake --build --preset linux-x64-release
   ```
   (Adjust the preset for your platform if needed — see `README.md` for options.)

That's it! This is a one-line change with no impact on SDK behavior.

### ✅ Acceptance Criteria

To merge a pull request for this issue:

- [ ] The `<cstdint>` include is added to `HbarUnit.h`
- [ ] No other files are modified
- [ ] No SDK behavior or public API changes
[...]

Issue #1124, opened 2026-02-04
Missing gtest in vcpkg.json


tech0priyanshu:
Description

1. While working with vcpkg integration, I noticed two related issues in the repository:
2. The invpkg.json file for gtest is missing.

tech0priyanshu:
1. yes 
2. ```Error:Could NOT find GTest``` rightnow  
3. Yeah  i mean `vcpkg.json`. yeah installing separate can be resolve by adding that in ```vcpkg```.
4. `vcpkg_installed` and `manifest-info.json` are vcpkg-generated artifacts. They’re not source, are machine-specific, and can be recreated from the manifest/lock files. it reduced noise for repo else i was seeing lot many change in my local.

tech0priyanshu:
Yeah fine 👍

PR #1125, opened 2026-02-04
Fix Linux build issues: add missing vcpkg artifacts and <cstdint> includes

### LINKED ISSUE #1123: Build failure on Linux: missing <cstdint> includes in public headers
Description

Building hiero-sdk-cpp on Linux (GCC/Clang) fails due to missing <cstdint> includes in public headers that use fixed-width integer types (uint8_t, uint32_t, uint64_t).

These headers compile on MSVC due to indirect includes but correctly fail on GCC/Clang, causing widespread compilation errors.

Affected Headers
Files directly use uint*_t types without including <cstdint>.

tech0priyanshu:
**Description**:
Fix Linux build failures related to vcpkg and missing headers

**Related issue(s)**:

Fixes #1123  #1124

**Notes for reviewer**:

**Checklist**


rwalworth (changes requested):
Thanks for working on this @tech0priyanshu! There are a few things we need to sort out:

**Issue #1124 — Unanswered questions:**
I asked some clarifying questions on issue #1124 that haven't been answered yet. Please respond to those questions before we proceed with the vcpkg.json and .gitignore changes. It's possible those changes aren't needed.

**Issue #1123 — Coordination with #1104:**
As I mentioned in issue #1123, `HbarUnit.h` is covered by issue #1104, which is marked as a "Good First Issue" and @Adityasakare has requested assignment. Please either:
- Exclude `HbarUnit.h` from this PR to let them complete that issue, OR
- Coordinate with @Adityasakare to let them know this PR will address it

**Unrelated formatting changes:**
I noticed `ContractId.h` and `PublicKey.h` have extensive formatting changes beyond the `<cstdint>` include — comment line wrapping, blank line removals, namespace closing comments, etc. These shouldn't be needed as clang-format checks are already passing in CI on `main`.

Please revert these formatting changes and only include the `<cstdint>` additions. This will make the diff easier to review and avoid potential merge conflicts.

**GPG Signing:**
The commits also need GPG signatures. Both have DCO sign-off (thanks!), but they're missing the `-S` flag. Once we sort out the above items, please follow the [Signing Requirements](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/docs/training/signing.md) to re-sign your commits.

Let me know if you have any questions!

rwalworth:
The coordination concern with #1104 is now resolved. @Adityasakare has decided to pass on the issue, and I've closed #1104 as a duplicate of #1123.

The `<cstdint>` fix in `HbarUnit.h` can proceed as part of this PR.

rwalworth (changes requested):
So the `<cstdint>`/`<string>` adds look good and match #1123. Two things need to be addressed before we can merge:

1. **Build failure**: The Linux workflow is failing when linking the unit and integration test executables (undefined reference to `testing::internal::MakeAndRegisterTestInfo`). Adding gtest to `vcpkg.json` means the tests must link against the vcpkg-provided GTest via the imported targets `GTest::gtest` and `GTest::gtest_main`. Right now the test CMakeLists use `gtest_main gtest`, which resolves to system libraries and causes the mismatch.

2. Please reply on #1124 to the open question about which CMake commands you're running and where you're seeing `vcpkg_installed/` and `manifest-info.json`. When using the preset workflow, those artifacts live under `build/` (which is already ignored), so we shouldn't add them to `.gitignore` until we know your workflow justifies it.

Once both are addressed, we can re-review and merge.

rwalworth (approved):
LGTM — the changes correctly address the Linux build issues: adding `<cstdint>` where fixed-width types are used (so Linux compilers get the right definitions) and adding the missing `gtest` dependency to `vcpkg.json` makes sense. I left one small nit below on `.gitignore`; nothing blocking. I'll run the workflows and will merge once they pass.

Congratulations on your first contribution to the Hiero C++ SDK @tech0priyanshu! Thank you for taking the time to contribute.

rwalworth:
Hey @tech0priyanshu, did you need any assistance with fixing the build?

Issue #1146, opened 2026-02-06
[Good First Issue]: Missing system dependencies & header hygiene checks on Ubuntu


tech0priyanshu:
#### Description

While setting up and building the project on Ubuntu, I encountered multiple issues caused by missing system-level dependencies and undocumented requirements. These can block new contributors during local setup.

This issue tracks **three related setup gaps** discovered during installation and verification.

---

###  Missing `zip` utility

Some build or packaging steps rely on the `zip` utility, which is not installed by default on a fresh Ubuntu system.

**Fix:**

```bash
sudo apt install zip
```

---

###  OpenSSL requires Linux kernel headers on Ubuntu

OpenSSL requires Linux kernel headers provided by the system package manager. On Ubuntu/Debian systems, this dependency is `linux-libc-dev`, which is not always present by default.

**Fix (Ubuntu/Debian):**

```bash
sudo apt install linux-libc-dev
```

**Reference:**

* Alpine: `apk add linux-headers`
* Ubuntu/Debian: `apt install linux-libc-dev`

Without this package, OpenSSL-related build steps fail.

---

###  Suggested Improvements

* Document required system dependencies (`zip`, `linux-libc-dev`) in the README or setup guide
* Optionally add a pre-build check to surface clearer errors when system dependencies are missing

---

### Steps to reproduce

Try to build in wsl  without this things

### Version

latest

### Operating system

Linux

rwalworth:
You're assigned @JarvisX-Z!

PR #1147, opened 2026-02-06
Add Linux dependency checks for zip utility and linux headers #1146

JarvisX-Z:
This PR adds system dependency checks for Linux environments. It ensures that required packages such as `zip` and `linux-libc-dev` are available before building the project.

## Changes Made
- Added check for `zip` utility using `find_program`
- Added check for Linux headers using `CheckIncludeFile`
- Displays clear error messages if dependencies are missing

## Why This Change?
This helps prevent build failures by notifying users about missing required dependencies earlier in the setup process.

### DIFF
--- a/src/sdk/main/include/PrivateKey.h
+++ b/src/sdk/main/include/PrivateKey.h
@@ -47,7 +47,7 @@ public:
    * Construct a PrivateKey object from a hex-encoded, DER-encoded key string.
    *
    * @param key The DER-encoded hex string from which to construct a PrivateKey.
-   * @return A pointer to an PrivateKey representing the input DER-encoded hex string.
+   * @return A pointer to a PrivateKey representing the input DER-encoded hex string.
    * @throws BadKeyException If the private key type (ED25519 or ECDSAsecp256k1) is unable to be determined or realized
    *                         from the input hex string.
    */
--- a/src/sdk/main/include/PublicKey.h
+++ b/src/sdk/main/include/PublicKey.h
@@ -45,7 +45,7 @@ public:
    * Construct a PublicKey object from a hex-encoded, DER-encoded key string.
    *
    * @param key The DER-encoded hex string from which to construct a PublicKey.
-   * @return A pointer to an PublicKey representing the input DER-encoded hex string.
+   * @return A pointer to a PublicKey representing the input DER-encoded hex string.
    * @throws BadKeyException If the public key type (ED25519 or ECDSAsecp256k1) is unable to be determined or realized
    *                         from the input hex string.
    */


SystemLibraries.cmake
```diff
@@ -8,6 +8,26 @@ if(NOT WIN32)
     set_target_properties(sys::resolv PROPERTIES IMPORTED_LOCATION ${RESOLV_LIBRARY})
 endif()
 
+ # Linux dependency checks
+ if(UNIX AND NOT APPLE)
```
rwalworth:
```suggestion
if(UNIX AND NOT APPLE)
```

JarvisX-Z:
Addressed the requested changes. Please review again.

rwalworth (dismissed):
LGTM! I will run the workflows and merge. Thank you @JarvisX-Z!

rwalworth (changes requested):
So it looks like this fix isn't going to work since the project only enables CXX (`project(... LANGUAGES CXX)`). So **C is not an enabled language**. When `check_include_file` runs, it uses that variable and tries to compile a `.c` file. CMake then errors with:

```text
Unknown extension ".c" for file .../CheckIncludeFile.c
try_compile() works only for enabled languages.  Currently these are:
  CXX
```

So the check fails **not** because linux-libc-dev is missing on the runner, but because we’re asking `try_compile` to build a C source in a CXX-only project. The "linux-libc-dev is required but not installed" message is misleading in this case — the configure fails before we’ve really tested for the header.

I see two options:

1. **Remove** the `set(CMAKE_TRY_COMPILE_SOURCE_EXTENSION c)` line (and leave `check_include_file` as-is). Then the check will use C++ (the only enabled language) and should pass on CI.
2. Go with the fix I recommended before: **replace** the `CheckIncludeFile` block with a `find_path` check for `linux/version.h` (e.g. under `/usr/include`). That avoids try_compile and the C/CXX issue entirely.

rwalworth (approved):
Thanks for the updates @JarvisX-Z — the zip and linux-libc-dev checks look good and match what we want for #1146. Two small indentation nits:

1. The first comment line in the new block (`# Linux dependency checks`) has a leading space before the `#` — in CMake, comment lines usually start with `#` in column 1
2. The `find_path`, `if(NOT LINUX_VERSION_HEADER)`, and its `endif()` aren’t indented, while the rest of the `if(UNIX AND NOT APPLE)` block uses 4 spaces.

You can fix those now if you’d like, or we can merge as-is and I’ll open a follow-up issue for the indentation cleanup. Your call! Running the workflows now and will merge once they pass.

PR #1279, opened 2026-03-25
feat: improve issue submission and triage workflow

rwalworth:
## Summary

This PR overhauls the issue submission and triage workflow for the Hiero C++ SDK. It replaces the five skill-level-specific issue submission templates with three problem-type templates (Bug, Feature, Task), adds a `/finalize` bot command that transitions triaged issues to ready-for-dev state, and adds maintainer and contributor documentation to support the new process.

**Key Changes:**
- Remove 5 skill-level issue templates; replace with 3 problem-type templates (Bug, Feature, Task)
- Add `/finalize` bot command that validates labels, updates issue title/body, and swaps status labels
- Extract shared helpers (`api.js`, `logger.js`) and refactor `assign.js` / `unassign.js` to use them
- Add maintainer triage guide and contributor issue-type guide
- Comprehensive snapshot-based tests for the finalize command (778+ lines)

## Motivation

Previously, issue submitters chose a skill-level template themselves. Contributors cannot reliably assess implementation complexity, so difficulty labels were inconsistently applied and the triage step was unclear. The new process separates concerns:

- **Submitters** describe the problem (Bug, Feature, or Task)
- **Maintainers** assess complexity, apply skill/priority/kind labels, then run `/finalize`
- **Contributors** pick up well-structured, clearly labeled issues via `/assign`

This makes the contribution experience more predictable for everyone.

---

### Issue Templates

Removed the five skill-level submission templates that asked contributors to self-assess complexity, which led to mislabeled issues:

| Removed | Reason |
|---------|--------|
| `01_good_first_issue_candidate.yml` | Skill assessment is now a maintainer responsibility |
| `02_good_first_issue.yml` | " |
| `03_beginner_issue.yml` | " |
| `04_intermediate_issue.yml` | " |
| `05_advanced_issue.yml` | " |

Added three problem-type templates that focus on describing the issue clearly:

| Added/Updated | Type | Purpose |
|---|---|---|
| `bug.yml` | Bug | Report broken or unexpected behavior |
| `feature.yml` | Feature | Request a new user-facing SDK capability |
| `task.yml` | Task (new) | Propose maintenance, refactoring, or improvement work |

All three templates now:
- Auto-apply `status: awaiting triage` (instead of a skill-level label)
- Include inline examples embedded as hidden HTML comments to guide submitters without cluttering the form
- Require structured fields: description, reproduction steps / proposed approach / implementation steps, expected/actual behavior, acceptance criteria

### `/finalize` Bot Command

[...]

PR #1290, opened 2026-03-28
feat: implement issue recommendation system for contributor progression

### LINKED ISSUE #1261: [Intermediate]: Issue recommendation system for contributor progression
## Problem :

Currently, there is no structured way to guide contributors to the next relevant issues after completing a contribution.

Contributors may finish an issue but lack visibility into what to work on next, especially as they progress from beginner to intermediate or advanced levels.

This becomes more challenging across multiple repositories, where issues are distributed and there is no consistent mechanism to recommend suitable next tasks based on contributor experience.

## Solution :

Propose introducing an issue recommendation system as part of the existing GitHub Actions automation, to suggest relevant issues to contributors based on their contribution history and difficulty progression.

The system would:

- Determine the difficulty level of the completed issue (e.g., good first issue, beginner, intermediate, advanced)
- Track contributor progress based on merged PRs

- Recommend issues based on a progression strategy:
  - Prefer one level higher (if the contributor is eligible)
  - Otherwise recommend issues at the same level
  - If no relevant issues are available, fall back to one level lower  
    - Exception: for beginner → good first issue, skip fallback since good first issues are intended for first-time contributors
  - If no matching issues are found in the current repository, optionally recommend issues from other repositories

- Support configurable behavior via labels and difficulty mappings

This approach ensures contributors are guided progressively while keeping the system flexible and reusable across repositories.

parvninama:
### Issue recommendation workflow on PR merge
- Triggers recommendation system when a PR is merged
- Ensures only valid contributor PRs are processed (non-bot, merged PRs)

### Linked issue resolution
- Added `resolveLinkedIssue` helper using `closingIssuesReferences` (GraphQL)
- When multiple issues are linked, selects the one with the highest skill level
- Uses the linked issue as the source of truth for labels
- Skips recommendation if no linked issue is found

### Difficulty-based progression system
- Introduced `SKILL_HIERARCHY` in constants (frozen):
  - good first issue → beginner → intermediate → advanced
- Determines completed issue level from labels (highest level wins for multi-label issues)
- Recommends issues using progression:
  - next level (primary)
  - same level (fallback)
  - lower level (final fallback, except Beginner → GFI)

### Issue filtering and discovery
- Fetches all `ready for dev` issues in a single API call (`per_page: 50`)
- Filters client-side by skill level to avoid GitHub Search API OR limitations
- Filters:
  - open issues only
  - unassigned issues
  - `ready for dev` label required
- Limits results to a small, relevant set (top 5)

### Error handling and resilience
- Differentiates:
  - `[]` → no issues found
  - `null` → API failure
- Posts error comment tagging maintainers on API failure
- Silent exit when no recommendations are available (avoids noise)

### Recommendation comment system
- Builds structured, readable comment with clickable issue links
- Mentions contributor directly

### Code structure improvements
- Moved `SKILL_HIERARCHY` to `helpers/constants.js` for shared use across `api.js` and `recommend-issues.js`
- Extracted linked issue resolution into reusable helper (`resolveLinkedIssue`)
- Kept API logic centralized in helpers
- Maintained separation between:
  - command logic (recommendation)
  - API utilities
  - workflow entry point

---

### Case 1 : Next level recommendation 

- Logs:
  - https://github.com/parvninama/hiero-sdk-cpp/actions/runs/23769338405/job/69256786093
- PR:
  - https://github.com/parvninama/hiero-sdk-cpp/pull/35

### Case 2: Same level recommendation 

- Logs:
  - https://github.com/parvninama/hiero-sdk-cpp/actions/runs/23769532965/job/69257419550
- PR:
  - https://github.com/parvninama/hiero-sdk-cpp/pull/36

### Case 3: Fallback level recommendation

- Logs:
  - https://github.com/parvninama/hiero-sdk-cpp/actions/runs/23769094183/job/69255977643
- PR: 
  - https://github.com/parvninama/hiero-sdk-cpp/pull/34

### Case 4: Multiple linked issues

- Logs:
  - https://github.com/parvninama/hiero-sdk-cpp/actions/runs/23770793343/job/69261534079
- PR:
  - https://github.com/parvninama/hiero-sdk-cpp/pull/43

### Case 5: No linked issues

- Logs:
  - https://github.com/parvninama/hiero-sdk-cpp/actions/runs/23770911024/job/69261913772
- PR:
  - https://github.com/parvninama/hiero-sdk-cpp/pull/44

### Case 6: No READY_FOR_DEV issues

- Logs:
  - https://github.com/parvninama/hiero-sdk-cpp/actions/runs/23771199309/job/69262851474
- PR:
  - https://github.com/parvninama/hiero-sdk-cpp/pull/45
---

## Linked Issue:
- Fixes #1261

### DIFF
--- a/src/sdk/main/include/TokenCancelAirdropTransaction.h
+++ b/src/sdk/main/include/TokenCancelAirdropTransaction.h
@@ -56,7 +56,7 @@ public:
    * Set the Pending Airdrops to be claimed
    *
    * @param pendingAirdrops The list of Pending Airdrop Id objects
-   * @return A reference to this TokenClaimAirdropTransaction with the newly-set pending airdrops.
+   * @return A reference to this TokenCancelAirdropTransaction with the newly-set pending airdrops.
    */
   TokenCancelAirdropTransaction& setPendingAirdrops(const std::vector<PendingAirdropId>& pendingAirdrops);
 


rwalworth (approved):
LGTM @parvninama - the tests look great and cover all the cases I asked for (boundary conditions, fallback skipping for beginner→GFI, API failure paths, and all the short-circuit cases in `handleRecommendIssues`). The `SKILL_HIERARCHY` extraction and `resolveLinkedIssue` helper are also well-structured.

Running the workflows now and will merge once they pass!

Congratulations on your first contribution to the Hiero C++ SDK @parvninama! 🎉 This was a well-designed feature end-to-end - thank you for sticking with it through the iterations!

rwalworth:
> > > @rwalworth All requested changes have been added
> > > One thing worth flagging on `resolveLinkedIssue` — when multiple issues are linked but none have a skill label, the current implementation returns the first fetched issue , Should it return `null` instead?
> > > Since this is a helper it may get reused in other contexts, and a stricter contract (no skill label → null) might be safer
> > > Happy to make that change if preferred, or track it as a follow-up
> > 
> > 
> > Good catch - and yes, I agree with your instinct here. Returning `null` when none of the linked issues have a skill label would be the cleaner contract.
> > The purpose of `resolveLinkedIssue` is to find the best issue for determining skill level. When no issue can provide a meaningful level, returning a "winner by default" (the first fetched) is technically correct but semantically misleading - the caller gets a truthy value but can't do anything useful with it. The `null` path through `bot-on-pr-close.js` already handles this case gracefully, so the end behavior is the same either way.
> > The fix is small:
> > 
> > * In the `reduce`, treat issues with no skill label (index `-1`) as losing to everything, including the initial accumulator value of `null`
> > * Update the test for "multiple issues, none with skill label" to assert `null` instead of the first fetched issue
> > 
> > This doesn't block the PR - the downstream code handles both paths correctly - so feel free to either make the change here or open a follow-up issue. Up to you!
> 
> Since this also requires updating the tests, I’ll open a follow-up issue and handle it there once this PR is merged to keep things focused. I’d be happy to take that up as well.
> 
> Also noted the points around getIssueSkillLevel having different behaviors across files and the repeated logger delegation pattern — both seem worth addressing as separate follow-ups.
> 
> I’d be interested in taking a look at those as well.

Sounds like a great plan @parvninama! So to fully break down the follow-up work:

1. **`resolveLinkedIssue` null contract** - update the `reduce` logic so that when no linked issue has a skill label, the function returns `null` rather than the first fetched issue. Update the existing test to cover this case.

2. **`getIssueSkillLevel` behavior divergence** - `assign.js` iterates forward (returns the lowest matching level) while `recommend-issues.js` iterates in reverse (returns the highest). They produce the same result today on single-label issues, but the name collision is a long-term trap. The fix is to either rename one or extract a shared helper with a single, well-defined contract.

3. **Logger delegation in `recommend-issues.js`** - open a follow-up issue to update `recommend-issues.js` to import `createDelegatingLogger` from `helpers/logger.js` instead of using the pattern inline. Note that #1279 is introducing that helper, so the follow-up PR should be opened after #1279 merges (which will hopefully be tomorrow).

Link back to this PR for context when you open the issues. This will be merged once we get an approval from @hiero-ledger/github-maintainers - looking forward to the follow-ups!

rwalworth:
Hey @parvninama - the PR is merged! You can go ahead and create those follow-up issues now. See my [previous comment](https://github.com/hiero-ledger/hiero-sdk-cpp/pull/1290#issuecomment-4189914458) for the issue breakdown (#1279 also merged, so the logger delegation one is unblocked too).

Thanks again for the great work on this one - looking forward to seeing more great contributions from you!

PR #1365, opened 2026-04-08
fix(bot): migrate to REST API and implement live state verification to resolve TOCTOU assignment race conditions

### LINKED ISSUE #1358: [Intermediate]: Bug: TOCTOU race condition in bot allows users to bypass assignment limits
### 🧩 Intermediate Friendly

> [!IMPORTANT]
> ### 🧭 About Intermediate Issues
>
> Intermediate Issues are a great next step for contributors who enjoy digging into the codebase and reasoning about how things work.
>
> These issues often:
> - Involve multiple related files or components  
> - Encourage investigation and understanding of existing behavior  
> - Leave room for thoughtful implementation choices  
> - Stay focused on a clearly defined goal  
>
> Other kinds of contributions — from beginner-friendly tasks to large system-level changes — are just as valuable and use different labels.

### Description of the Issue

Hey team! While looking through the bot infrastructure to understand the contribution workflow, I noticed a classic Time-Of-Check to Time-Of-Use (TOCTOU) race condition in the /assign command logic. This flaw currently allows contributors to bypass the MAX_OPEN_ASSIGNMENTS limit (and the GFI limit).

The issue stems from the countAssignedIssues function inside .github/scripts/commands/assign.js. It relies on the github.graphql API to run a search(type: ISSUE). The problem is that GitHub's Search API is powered by Elasticsearch, which has a well-known indexing lag (often taking 30–60 seconds to catch up).

Because the bot relies on this eventually-consistent index to check eligibility, the pre-assignment check is acting on stale data during concurrent requests.

### Steps to Reproduce

1. Find 3 to 5 unassigned issues that have the status: ready for dev label.
2. Open them in separate browser tabs.
3. Rapidly comment /assign on all of them within a few seconds of each other.
4. Because the GitHub GraphQL search index hasn't updated yet, the bot reads count = 0 for every single concurrent run.
5. You are successfully assigned to all 5 issues, completely bypassing the limit of 2.

### Expected Behavior

The bot should reliably cap the user at the configured MAX_OPEN_ASSIGNMENTS, gracefully rejecting the concurrent requests (3rd, 4th, etc.) even if they are fired at the exact same time.

### Actual Behavior

The pre-check reads stale Elasticsearch data. The bot blindly processes the concurrent requests successfully, leaving the user assigned to far more issues than the limit allows.

### Environment

- SDK version: N/A (GitHub Actions Bot Infrastructure)
- Operating system: GitHub Actions / Node.js
- Compiler and version: N/A
- Hiero network: N/A

### Acceptance Criteria

- [ ] The described behavior is no longer reproducible
- [ ] A regression test is added that fails before the fix and passes after
- [ ] No unrelated behavior or API changes are introduced
- [ ] All existing tests continue to pass

---

### 📋 Step-by-Step Contribution Guide

To help keep contributions consistent and easy to review, we recommend following these steps:

- [ ] Wait for assignment

Read [Workflow Guide](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/docs/training/workflow.md) for step-by-step workflow guidance.
Read [README.md](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/README.md) for setup instructions.

See the [Signing Guide](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/docs/training/signing.md).

### 🤔 Additional Information

I looked into a few ways to solve this. Using a YAML-level concurrency lock (e.g., concurrency: group: bot-assign-${{ github.event.repository.name }}) would work, but it would serialize all issue comments globally and heavily bottleneck repo velocity.

A much better architectural fix is to:

1. Swap the GraphQL search out for the REST API (issues.listForRepo), which hits strongly-consistent SQL replicas and reduces the vulnerability window from 30 seconds down to ~100ms.

2. Implement an Optimistic Concurrency Control (OCC) pattern in assignAndFinalize. We can assign the user, wait ~500ms for REST replica sync, and then fetch the global count again. If the race condition was hit and the limit is breached, we just invoke removeAssignees and post a rollback comment.
I actually have a branch ready with this exact OCC REST implementation! Let me know if you agree with the approach and I can open the PR to fix this immediately.

darshit2308:
**Description**:

Migrate the /assign bot from GraphQL-based counting to REST-based counting, and implement the agreed two-part mitigation for stale-count assignment-limit bypass and same-issue queue races.

* Replace GraphQL assignment counting with REST issues.listForRepo counting.
* Filter out pull requests from the issues endpoint response before counting.
* Exclude status: blocked issues from open-assignment limit checks when no explicit label filter is requested.
* Add threshold-based early exit in counting paths to reduce unnecessary pagination and API calls.
* Keep workflow concurrency scoped per issue to prevent same-issue double-assignment races.
* Add a fresh pre-write issues.get check before addAssignees to handle queued stale-payload scenarios.
* Refactor assign bot test mocks to REST-oriented naming and behavior.
* Add/keep regression scenarios for queue snatch and requester already assigned in fresh state.
* Remove post-write OCC rollback paths so the implementation matches the agreed design in issue #1358.

**Related issue(s)**:

Fixes #1358

**Notes for reviewer**:

As discussed in issue #1358, this change follows the agreed strategy:

1. Count consistency race:
   REST-based counting removes GraphQL search-index lag from assignment-limit gates.

2. Same-issue execution race:
   Per-issue workflow serialization plus fresh pre-write issues.get handles same-issue queue collisions caused by stale event payloads.

No post-write rollback path is used in the final implementation.

Local bot test suites pass, including the updated assign integration suite (31/31).

**Checklist**

- [x] Documented (code comments, etc.)

### DIFF
--- a/.github/scripts/tests/test-assign-bot.js
+++ b/.github/scripts/tests/test-assign-bot.js
@@ -27,6 +27,7 @@ function createMockGithub(options = {}) {
     assignShouldFail = false,
     removeLabelShouldFail = false,
     addLabelShouldFail = false,
+    reactionShouldFail = false,
     issueGetShouldFail = false,
     issueAssignees = null,
   } = options;
@@ -45,6 +46,11 @@ function createMockGithub(options = {}) {
     rest: {
       reactions: {
         createForIssueComment: async (params) => {
+          if (reactionShouldFail) {
+            throw new Error(
+              "Simulated reaction failure: comment not found (404)",
+            );
+          }
           calls.reactions.push({
             commentId: params.comment_id,
             content: params.content,
@@ -1247,26 +1253,27 @@ Error details: Failed to remove 'status: ready for dev': Simulated remove label 
   // ---------------------------------------------------------------------------
 
   {
-    name: 'Abort - Triggering Comment Deleted',
-    description: 'Bot aborts /assign flow when acknowledgeComment fails (comment deleted)',
+    name: "Abort - Triggering Comment Deleted",
+    description:
+      "Bot aborts /assign flow when acknowledgeComment fails (comment deleted)",
     context: {
-      eventName: 'issue_comment',
+      eventName: "issue_comment",
       payload: {
         issue: {
           number: 400,
           assignees: [],
           labels: [
-            { name: 'status: ready for dev' },
-            { name: 'skill: good first issue' },
+            { name: "status: ready for dev" },
+            { name: "skill: good first issue" },
           ],
         },
         comment: {
           id: 4001,
-          body: '/assign',
-          user: { login: 'deleted-comment-user', type: 'User' },
+          body: "/assign",
+          user: { login: "deleted-comment-user", type: "User" },
         },
       },
-      repo: { owner: 'hiero-ledger', repo: 'hiero-sdk-cpp' },
+      repo: { owner: "hiero-ledger", repo: "hiero-sdk-cpp" },
     },
     githubOptions: { reactionShouldFail: true },
     expectedAssignee: null,


darshit2308:
Hi @rwalworth, thank you for the careful review,  you're right to flag that.
The version I originally pushed was an earlier iteration that still included the OCC rollback logic and removeAssignees. That was a mistake on my part ,  I had gone through several rounds of revision locally and submitted the wrong branch. I completely agree the symmetric rollback trap makes OCC incorrect here without atomic primitives, which is exactly why we settled on the two-pronged approach.

I've pushed an updated version that removes all OCC logic (removeAssignees, the post-write re-count, and the rollback comment) entirely. The implementation now strictly follows what we agreed on in #1358:

1) countAssignedIssues migrated to listForRepo REST with client-side blocked-issue filtering,  I also added pull-request filtering since the issues endpoint returns PRs, and an early-exit threshold parameter to avoid fetching more pages than necessary.
2) Fresh issues.get() pre-flight inside assignAndFinalize() before addAssignees(), relying on the existing per-issue concurrency key to serialize same-issue runs.

So the current flow is exactly the agreed two-pronged design, with no post-write rollback path.

Sorry for the noise , let me know if you'd like me to clarify anything in the updated PR.

darshit2308:
Hi @rwalworth , thank you again for the review and guidance. I have resolved the merge conflicts.
One non-blocking follow-up I noticed while validating the final flow: ready and skill labels are validated from the event payload, while the pre-write fresh fetch currently rechecks assignees.
That means if labels change between comment creation and execution, there is still a stale-label timing edge. This is outside the assignment-limit TOCTOU scope in 1358 and does not affect this PR’s fix.
If you agree, I can open a separate issue and follow-up PR for that hardening after this one merges, or I can make changes in this PR itself and do the correction.

darshit2308:
yes sure, I will open an issue regarding this.

andrewb1269 (approved):
Review and approve the workflow file updates.

PR #1424, opened 2026-04-14
feat: Implement updateContract in TCK contract service

### LINKED ISSUE #1377: [Intermediate]: Implement `updateContract` in the TCK contract service
## 🛠️ Intermediate Issue

This issue is a great fit for contributors who are ready to own a solution end-to-end.

Intermediate Issues involve exploring existing patterns, reasoning about trade-offs, and choosing an implementation approach with confidence. You'll navigate multiple related files and explain your decisions in review — the implementation details are yours to work out.

> [!IMPORTANT]
> **Prerequisite:** You must have completed at least **3 Beginner Issues** in this repository before being assigned. The `/assign` bot enforces this automatically.

---

## 👾 Description of the Task

The Hiero C++ SDK TCK server does not yet implement the `updateContract` JSON-RPC method. This method is required by the TCK test suite to verify that the SDK can correctly update smart contract properties on the Hedera network.

The TCK server exposes a JSON-RPC interface that maps incoming method calls to SDK transactions and queries. The contract service scaffolding (`ContractService.h`, `ContractService.cc`) either already exists (from the `createContract` issue) or needs to be created following the same pattern as `AccountService`.

Relevant files to create:
```
src/tck/include/contract/params/UpdateContractParams.h
```

Relevant files to modify:
```
src/tck/include/contract/ContractService.h
src/tck/src/contract/ContractService.cc
src/tck/src/TckServer.cc
```

Reference SDK class:
```
src/sdk/main/include/ContractUpdateTransaction.h
```

---

## 💡 Proposed Approach

Follow the existing service pattern established by `AccountService::updateAccount` and `TokenService::updateToken` — both are good references since they also update optional fields on an existing entity.

1. Create `UpdateContractParams.h` with an `adl_serializer` specialization to deserialize the JSON request
2. Add the `updateContract` function declaration to `ContractService.h`
3. Implement `updateContract` in `ContractService.cc` using `ContractUpdateTransaction`, only setting fields whose optional values are present
4. Register the method in `TckServer.cc` and add the explicit template instantiation

If `ContractService.h` and `ContractService.cc` do not yet exist, create them first following the pattern of `AccountService.h` and `AccountService.cc`, and ensure `ContractService.cc` is listed in `src/tck/CMakeLists.txt`.

The response should return only the transaction `status` string.

---

## 👩‍💻 Implementation Steps

- [ ] Check whether `src/tck/include/contract/ContractService.h` and `src/tck/src/contract/ContractService.cc` already exist
  - If not, create them following the pattern of `AccountService.h` and `AccountService.cc`, and add `src/contract/ContractService.cc` to `src/tck/CMakeLists.txt`
- [ ] Read `src/tck/include/account/params/UpdateAccountParams.h` to understand the update params pattern
- [ ] Create `src/tck/include/contract/params/UpdateContractParams.h` with the following optional fields:
  - `mContractId` (`std::optional<std::string>`) — JSON key `"contractId"`
  - `mAdminKey` (`std::optional<std::string>`) — JSON key `"adminKey"`
  - `mExpirationTime` (`std::optional<std::string>`) — JSON key `"expirationTime"` (seconds since epoch as string)
  - `mAutoRenewAccountId` (`std::optional<std::string>`) — JSON key `"autoRenewAccountId"`
  - `mAutoRenewPeriod` (`std::optional<std::string>`) — JSON key `"autoRenewPeriod"` (seconds as string)
  - `mMemo` (`std::optional<std::string>`) — JSON key `"memo"`
  - `mMaxAutomaticTokenAssociations` (`std::optional<int32_t>`) — JSON key `"maxAutomaticTokenAssociations"`
  - `mStakedAccountId` (`std::optional<std::string>`) — JSON key `"stakedAccountId"`
  - `mStakedNodeId` (`std::optional<std::string>`) — JSON key `"stakedNodeId"`
  - `mDeclineStakingReward` (`std::optional<bool>`) — JSON key `"declineStakingReward"`
  - `mCommonTxParams` (`std::optional<CommonTransactionParams>`) — JSON key `"commonTransactionParams"`
- [ ] Add `updateContract` declaration to `src/tck/include/contract/ContractService.h`
- [ ] Implement `updateContract` in `src/tck/src/contract/ContractService.cc`:
  - Construct a `ContractUpdateTransaction`
  - Set `setGrpcDeadline(SdkClient::DEFAULT_TCK_REQUEST_TIMEOUT)`
  - Apply each optional param only if present — investigate how `ContractUpdateTransaction` handles key updates (check whether a key string requires parsing via the key service utilities)
  - Apply `mCommonTxParams` if present via `fillOutTransaction`
  - Execute and get the receipt; return `{ "status": ... }`
- [ ] Add includes, `addMethod` call, and explicit template instantiation for `UpdateContractParams` to `src/tck/src/TckServer.cc`
- [ ] Build the project:
  ```bash
  cmake --preset linux-x64-debug && cmake --build --preset linux-x64-debug
  ```
- [ ] Confirm the TCK server binary compiles and the new method is reachable
- [ ] Open a pull request with the change

---

## ✔️ Acceptance Criteria

- [ ] `UpdateContractParams.h` is created following the established params pattern
- [ ] `updateContract` is declared in `ContractService.h` and implemented in `ContractService.cc`
- [ ] `updateContract` is registered in `TckServer.cc` with its explicit template instantiation
- [ ] The TCK server compiles without errors or warnings introduced by this change
- [ ] All optional params are handled correctly; missing params do not cause crashes or spurious field resets
- [ ] No unrelated files are modified

---

## 📋 Step-by-Step Contribution Guide

- [ ] Wait for assignment

Read [Workflow Guide](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/docs/training/workflow.md) for step-by-step workflow guidance. Read [README.md](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/README.md) for setup instructions.

❗ Pull requests **cannot be merged** without `S` and `s` signed commits. See the [Signing Guide](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/docs/training/signing.md).

---

## 🤔 Additional Information

- Reference implementation pattern (update): `src/tck/src/account/AccountService.cc` (`updateAccount`), `src/tck/src/token/TokenService.cc` (`updateToken`)
- Reference params pattern: `src/tck/include/account/params/UpdateAccountParams.h`
- TCK test file (in the hiero-sdk-tck repo): `src/tests/contract-service/test-contract-update-transaction.ts`
- SDK transaction class: `src/sdk/main/include/ContractUpdateTransaction.h`

If you have questions while working on this issue, feel free to ask! [Hiero-SDK-C++ Discord](https://discord.com/channels/905194001349627914/1337424839761465364)

Omswastik-11:
Closes #1377

### DIFF
--- a/src/tck/src/contract/ContractService.cc
+++ b/src/tck/src/contract/ContractService.cc
@@ -53,6 +53,71 @@ std::string stripHexPrefix(std::string hex)
 
 namespace Hiero::TCK::ContractService
 {
+namespace
+{
+void applyUpdateContractCoreFields(const UpdateContractParams& params, ContractUpdateTransaction& transaction)
+{
+  if (params.mContractId.has_value())
+  {
+    transaction.setContractId(ContractId::fromString(params.mContractId.value()));
+  }
+
+  if (params.mAdminKey.has_value())
+  {
+    transaction.setAdminKey(KeyService::getHieroKey(params.mAdminKey.value()));
+  }
+
+  if (params.mExpirationTime.has_value())
+  {
+    transaction.setExpirationTime(
+      std::chrono::system_clock::from_time_t(0) +
+      std::chrono::seconds(internal::EntityIdHelper::getNum<int64_t>(params.mExpirationTime.value())));
+  }
+
+  if (params.mAutoRenewAccountId.has_value())
+  {
+    transaction.setAutoRenewAccountId(AccountId::fromString(params.mAutoRenewAccountId.value()));
+  }
+}
+
+void applyUpdateContractRenewalFields(const UpdateContractParams& params, ContractUpdateTransaction& transaction)
+{
+  if (params.mAutoRenewPeriod.has_value())
+  {
+    transaction.setAutoRenewPeriod(
+      std::chrono::seconds(internal::EntityIdHelper::getNum<int64_t>(params.mAutoRenewPeriod.value())));
+  }
+
+  if (params.mMemo.has_value())
+  {
+    transaction.setContractMemo(params.mMemo.value());
+  }
+
+  if (params.mMaxAutomaticTokenAssociations.has_value())
+  {
+    transaction.setMaxAutomaticTokenAssociations(params.mMaxAutomaticTokenAssociations.value());
+  }
+}
+
+void applyUpdateContractStakingFields(const UpdateContractParams& params, ContractUpdateTransaction& transaction)
+{
+  if (params.mStakedAccountId.has_value())
+  {
+    transaction.setStakedAccountId(AccountId::fromString(params.mStakedAccountId.value()));
+  }
+
+  if (params.mStakedNodeId.has_value())
+  {
+    transaction.setStakedNodeId(internal::EntityIdHelper::getNum<int64_t>(params.mStakedNodeId.value()));
+  }
+
+  if (params.mDeclineStakingReward.has_value())
+  {
+    transaction.setDeclineStakingReward(params.mDeclineStakingReward.value());
+  }
+}
+} // namespace
+
 //-----
 nlohmann::json createContract(const CreateContractParams& params)
 {
@@ -181,58 +246,9 @@ nlohmann::json updateContract(const UpdateContractParams& params)
   ContractUpdateTransaction contractUpdateTransaction;
   contractUpdateTransaction.setGrpcDeadline(SdkClient::DEFAULT_TCK_REQUEST_TIMEOUT);
 
-  if (params.mContractId.has_value())
-  {
-    contractUpdateTransaction.setContractId(ContractId::fromString(params.mContractId.value()));
-  }
-
-  if (params.mAdminKey.has_value())
-  {
-    contractUpdateTransaction.setAdminKey(KeyService::getHieroKey(params.mAdminKey.value()));
-  }
-
-  if (params.mExpirationTime.has_value())
-  {
-    contractUpdateTransaction.setExpirationTime(
-      std::chrono::system_clock::from_time_t(0) +
-      std::chrono::seconds(internal::EntityIdHelper::getNum<int64_t>(params.mExpirationTime.value())));
-  }
-
-  if (params.mAutoRenewAccountId.has_value())
-  {
-    contractUpdateTransaction.setAutoRenewAccountId(AccountId::fromString(params.mAutoRenewAccountId.value()));
-  }
-
-  if (params.mAutoRenewPeriod.has_value())
-  {
-    contractUpdateTransaction.setAutoRenewPeriod(
-      std::chrono::seconds(internal::EntityIdHelper::getNum<int64_t>(params.mAutoRenewPeriod.value())));
-  }
-
-  if (params.mMemo.has_value())
-  {
-    contractUpdateTransaction.setContractMemo(params.mMemo.value());
-  }
-
-  if (params.mMaxAutomaticTokenAssociations.has_value())
-  {
-    contractUpdateTransaction.setMaxAutomaticTokenAssociations(params.mMaxAutomaticTokenAssociations.value());
-  }
-
-  if (params.mStakedAccountId.has_value())
-  {
-    contractUpdateTransaction.setStakedAccountId(AccountId::fromString(params.mStakedAccountId.value()));
-  }
-
-  if (params.mStakedNodeId.has_value())
-  {
-    contractUpdateTransaction.setStakedNodeId(internal::EntityIdHelper::getNum<int64_t>(params.mStakedNodeId.value()));
-  }
-
-  if (params.mDeclineStakingReward.has_value())
-  {
-    contractUpdateTransaction.setDeclineStakingReward(params.mDeclineStakingReward.value());
-  }
+  applyUpdateContractCoreFields(params, contractUpdateTransaction);
+  applyUpdateContractRenewalFields(params, contractUpdateTransaction);
+  applyUpdateContractStakingFields(params, contractUpdateTransaction);
 
   if (params.mCommonTxParams.has_value())
   {
--- a/src/tck/src/contract/ContractService.cc
+++ b/src/tck/src/contract/ContractService.cc
@@ -179,68 +179,40 @@ nlohmann::json deleteContract(const DeleteContractParams& params)
 nlohmann::json updateContract(const UpdateContractParams& params)
 {
   ContractUpdateTransaction contractUpdateTransaction;
-  contractUpdateTransaction.setGrpcDeadline(SdkClient::DEFAULT_TCK_REQUEST_TIMEOUT);
+  auto& tx = contractUpdateTransaction;
+  tx.setGrpcDeadline(SdkClient::DEFAULT_TCK_REQUEST_TIMEOUT);
 
-  if (params.mContractId.has_value())
-  {
-    contractUpdateTransaction.setContractId(ContractId::fromString(params.mContractId.value()));
-  }
+  const auto applyIfSet = [](const auto& value, const auto& apply) {
+    if (value.has_value())
+    {
+      apply(value.value());
+    }
+  };
 
-  if (params.mAdminKey.has_value())
-  {
-    contractUpdateTransaction.setAdminKey(KeyService::getHieroKey(params.mAdminKey.value()));
-  }
-
-  if (params.mExpirationTime.has_value())
-  {
-    contractUpdateTransaction.setExpirationTime(
-      std::chrono::system_clock::from_time_t(0) +
-      std::chrono::seconds(internal::EntityIdHelper::getNum<int64_t>(params.mExpirationTime.value())));
-  }
-
-  if (params.mAutoRenewAccountId.has_value())
-  {
-    contractUpdateTransaction.setAutoRenewAccountId(AccountId::fromString(params.mAutoRenewAccountId.value()));
-  }
-
-  if (params.mAutoRenewPeriod.has_value())
-  {
-    contractUpdateTransaction.setAutoRenewPeriod(
-      std::chrono::seconds(internal::EntityIdHelper::getNum<int64_t>(params.mAutoRenewPeriod.value())));
-  }
-
-  if (params.mMemo.has_value())
-  {
-    contractUpdateTransaction.setContractMemo(params.mMemo.value());
-  }
-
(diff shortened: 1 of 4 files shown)


src/tck/src/contract/ContractService.cc
```diff
@@ -2,20 +2,95 @@
...
+    { transaction.setMaxAutomaticTokenAssociations(maxAutomaticTokenAssociations); });
+
+  applyOptionalValue<std::string>(params.mStakedAccountId,
+                                  [&transaction](const std::string& stakedAccountId)
+                                  { transaction.setStakedAccountId(AccountId::fromString(stakedAccountId)); });
+
+  applyOptionalValue<std::string>(
+    params.mStakedNodeId,
+    [&transaction](const std::string& stakedNodeId)
+    { transaction.setStakedNodeId(Hiero::internal::EntityIdHelper::getNum<uint64_t>(stakedNodeId)); });
```
rwalworth: **issue (blocking):** `getNum<uint64_t>` will throw `std::invalid_argument` when the TCK sends `stakedNodeId: "-100"` (TCK test "Staked ID #8"), because `std::from_chars` cannot parse negative values into `uint64_t`. This exception bubbles up and returns `INTERNAL_ERROR` to the TCK - but the test expects `INVALID_STAKING_ID`.

Change `uint64_t` to `int64_t` so the negative value is parsed successfully and passed to `setStakedNodeId`, where the network returns the expected `INVALID_STAKING_ID`:

```cpp
applyOptionalValue<std::string>(
  params.mStakedNodeId,
  [&transaction](const std::string& stakedNodeId)
  { transaction.setStakedNodeId(Hiero::internal::EntityIdHelper::getNum<int64_t>(stakedNodeId)); });
```

This is the same pattern used in `AccountService.cc` (line 561): `getNum<int64_t>` returns `-100` → implicitly cast to `uint64_t` when passed to `setStakedNodeId(const uint64_t&)` → the network receives a very large number and rejects it with `INVALID_STAKING_ID`. The implicit cast to `uint64_t` is expected behavior here.

src/tck/src/contract/ContractService.cc
```diff
@@ -2,20 +2,95 @@
...
+namespace
+{
+template<typename T, typename Setter>
+void applyOptionalValue(const std::optional<T>& value, Setter setter)
+{
+  if (value.has_value())
+  {
+    setter(value.value());
+  }
+}
```
rwalworth: **question (blocking):** What was the reason for introducing the `applyOptionalValue` helper and a separate `applyUpdateContractFields` function? Every other service in the codebase - `AccountService::updateAccount`, `TokenService::updateToken`, `ContractService::createContract`, `ContractService::deleteContract` - uses direct `if (params.xxx.has_value())` checks inline in the method body, with no separate helper. Did Codacy or something else flag the direct pattern?

Unless there's a specific reason, please refactor to match the established convention:

```cpp
if (params.mContractId.has_value())
{
  contractUpdateTransaction.setContractId(ContractId::fromString(params.mContractId.value()));
}

if (params.mAdminKey.has_value())
{
  contractUpdateTransaction.setAdminKey(KeyService::getHieroKey(params.mAdminKey.value()));
}
// ... etc.
```

rwalworth (changes requested):
Thanks for the contribution @Omswastik-11! The overall structure is solid - all 11 params from the TCK spec are present, JSON keys match exactly, and the registration in `TckServer.cc` is correct. I found one issue in the `stakedNodeId` handling that would cause a TCK test to fail, and a style note below. Once that's addressed, we should be good to go!

rwalworth (changes requested):
Thanks for the update @Omswastik-11! The `int64_t` fix for `stakedNodeId` is correct - that addresses the first comment. We're getting closer!

One item from my previous review is still outstanding: all three helper functions (`applyUpdateContractCoreFields`, `applyUpdateContractRenewalFields`, `applyUpdateContractStakingFields`) need to be removed and their contents inlined directly into `updateContract()` to match the codebase convention. Once this is fixed I'll run the workflows and merge!

src/tck/src/contract/ContractService.cc
```diff
@@ -171,6 +173,50 @@ nlohmann::json deleteContract(const DeleteContractParams& params)
...
+  ContractUpdateTransaction contractUpdateTransaction;
+  auto& tx = contractUpdateTransaction;
+  tx.setGrpcDeadline(SdkClient::DEFAULT_TCK_REQUEST_TIMEOUT);
+
+  const auto applyIfSet = [](const auto& value, const auto& apply) {
+    if (value.has_value())
+    {
+      apply(value.value());
+    }
+  };
```
rwalworth: **issue (blocking):** The `applyIfSet` lambda is still an abstraction over the `has_value()` checks rather than direct `if` blocks - same concern as before, just local instead of a separate function.

I can see what you're going for here - eliminating the repetitive `has_value()` / `.value()` boilerplate is a reasonable idea. But introducing this pattern in a single method while every other method in the TCK codebase (`createContract`, `deleteContract`, `AccountService::updateAccount`, `TokenService::updateToken`, etc.) uses direct `if` blocks creates an inconsistency that's harder to maintain. If this abstraction is worth adopting, it should be a codebase-wide refactor across all TCK service methods, proposed as a separate issue.

For this PR, please remove the `auto& tx` alias and replace all `applyIfSet(...)` calls with direct `if` blocks to match the established pattern:

```cpp
if (params.mContractId.has_value())
{
  contractUpdateTransaction.setContractId(ContractId::fromString(params.mContractId.value()));
}

if (params.mAdminKey.has_value())
{
  contractUpdateTransaction.setAdminKey(KeyService::getHieroKey(params.mAdminKey.value()));
}

if (params.mExpirationTime.has_value())
{
  contractUpdateTransaction.setExpirationTime(
    std::chrono::system_clock::from_time_t(0) +
    std::chrono::seconds(internal::EntityIdHelper::getNum<int64_t>(params.mExpirationTime.value())));
}

// ... and so on for each remaining field
```

Once this is updated I'll run the workflows and merge!

rwalworth (changes requested):
Thanks for the update @Omswastik-11! Good progress - the three separate helper functions are gone and the `int64_t` fix is correct. Just one more thing below before I can merge!

rwalworth (approved):
LGTM! Nice work working through those iterations @Omswastik-11 - all the feedback from the previous rounds has been addressed. Running the workflows now and will merge once they pass!

src/tck/src/contract/ContractService.cc
```diff
@@ -15,6 +16,7 @@
 #include <ContractCallQuery.h>
 #include <ContractCreateTransaction.h>
 #include <ContractDeleteTransaction.h>
+#include <ContractUpdateTransaction.h>
```
rwalworth: **chore (blocking):** `<ContractUpdateTransaction.h>` needs to move - the clang-format lint check sorts includes alphabetically, and "Update" (U) comes after "FunctionResult" (F), "Id" (I), "Info" (I), and "InfoQuery" (I). Move it to after `<ContractInfoQuery.h>`:

```cpp
#include <ContractInfoQuery.h>
#include <ContractUpdateTransaction.h>
#include <FileId.h>
```

You can also just run `clang-format -i src/tck/src/contract/ContractService.cc` from the repo root and it'll fix it automatically.

rwalworth (changes requested):
The clang-format lint check is failing because `<ContractUpdateTransaction.h>` is in the wrong include position in `ContractService.cc`. Left a comment below with the exact fix. Once that's pushed, I'll run the workflows and merge!

rwalworth (changes requested):
Thanks for the update @Omswastik-11! The include ordering fix is correct - `<ContractUpdateTransaction.h>` is now in the right position.

Two things to address:

**1. Clang-format failures (existing return statements)**

The `clang-format -i` run also reformatted the return initializer lists in pre-existing functions (`createContract`, `deleteContract`, `executeContract`, `contractByteCodeQuery`, `contractCallQuery`). Your local clang-format produced different output than CI's clang-format-17.0.6, which is why those 21 violations are showing up. The [Formatting Guide](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/docs/training/workflow.md#6-ensure-code-is-formatted) has the exact commands - note that on macOS Homebrew installs the binary as `clang-format` (not `clang-format-17`), which can produce slightly different output than the Linux CI runner.

Please revert those return statement changes back to the original - the only diff in `ContractService.cc` should be the new includes, the helper functions, and `updateContract()` itself. Also make sure the new `updateContract` return uses the same no-spaces style as `executeContract`: `{"status", gStatusToString.at(txReceipt.mStatus)}`.

**2. Helper functions**

I understand Codacy flagged the complexity when these were inlined. That said, if the `applyIfSet` abstraction is worth adopting, it should be applied consistently across all TCK service update methods - not just `updateContract`. Please open a follow-up issue for that and remove the three helper functions here so the implementation matches the rest of the codebase like I mentioned before.

Once both of those are addressed I'll run the workflows and merge!

Omswastik-11:
> 2. Helper functions
> 
> I understand Codacy flagged the complexity when these were inlined. That said, if the applyIfSet abstraction is worth adopting, it should be applied consistently across all TCK service update methods - not just updateContract. Please open a follow-up issue for that and remove the three helper functions here so the implementation matches the rest of the codebase like I mentioned before.

It fails the codacy complexity .

rwalworth:
> > 2. Helper functions
> > 
> > I understand Codacy flagged the complexity when these were inlined. That said, if the applyIfSet abstraction is worth adopting, it should be applied consistently across all TCK service update methods - not just updateContract. Please open a follow-up issue for that and remove the three helper functions here so the implementation matches the rest of the codebase like I mentioned before.
> 
> It fails the codacy complexity .

**note:** Codacy passing isn't a merge requirement - it's a quality signal, not a gate. The flags are a known trade-off of the established pattern and we'll accept them here.

That said, the `applyIfSet` abstraction you introduced earlier isn't a bad idea - it just needs to be applied consistently across all TCK service methods rather than in one place. If you're interested in a follow-up, please open an issue proposing the pattern codebase-wide and we can evaluate it there!

rwalworth (approved):
LGTM! Everything looks good and all the feedback from the previous rounds has been addressed. Running the workflows now and will merge once they pass!

Omswastik-11:
Thanks for clarifying I will create a follow up issue

PR #1507, opened 2026-04-24
feat: re-enable Dynamic Address Book (DAB) e2e tests 

### LINKED ISSUE #1004: [Intermediate]: Re-enable Dynamic Address Book (DAB) E2E Tests
## 🧩 Intermediate Friendly

---

## 🐞 Problem Description

The SDK is integrated with **solo-action** in its CI workflows. The Dynamic Address Book (DAB) end-to-end tests are currently disabled and need to be investigated, fixed, and re-enabled.

This task requires investigation across multiple areas: test code, SDK code, CI workflow configuration, and solo-action integration.

### What is Dynamic Address Book (DAB)?

The Dynamic Address Book ([HIP-869](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-869.md)) changed Hiero's manually administered address book into an HAPI-managed configuration that can be updated via signed transactions without restarting consensus nodes.

DAB introduced the `AddressBookService` with three transaction types:

| Transaction | Description | Required Signatures |
|-------------|-------------|---------------------|
| `NodeCreateTransaction` | Add a new consensus node to the network | Council signature AND admin key |
| `NodeUpdateTransaction` | Modify an existing node's properties | Admin key only |
| `NodeDeleteTransaction` | Remove a node from the network | Admin key OR council signature |

These transactions allow node operators to manage:
- Node IP addresses and ports
- gRPC service endpoints
- Node descriptions
- Public keys and certificates
- Node account IDs

### Current Problem

The DAB e2e tests are currently disabled in the SDK's CI workflows. These tests need to be:
1. Investigated to understand why they were disabled
2. Fixed if there are underlying bugs
3. Re-enabled in the CI workflow

---

## 💡 Expected Outcome

The DAB e2e tests should be fully functional and enabled in CI. This involves:

- Identifying which DAB tests are disabled and understanding why
- Fixing any bugs in the test code or SDK code that cause failures
- Ensuring tests pass consistently with the solo-action integration
- Re-enabling the tests in the appropriate CI workflow files

---

### Investigation Areas

This task requires exploring multiple areas to understand the root cause:

1. **Test code**: The tests themselves may have bugs or outdated assumptions
2. **SDK code**: The DAB transaction implementations may have issues
3. **CI workflow configuration**: Tests may be incorrectly excluded or configured
4. **Solo-action integration**: The local network setup may need adjustments for DAB

### Suggested Approach

- [ ] Locate the disabled tests and understand how they were disabled (skipped, commented out, excluded from CI, etc.)
- [ ] Review git history to understand when and why tests were disabled
- [ ] Run the tests locally to reproduce failures
- [ ] Investigate root causes across test code, SDK code, and CI configuration
- [ ] Fix identified issues in tests and/or SDK code
- [ ] Verify tests pass consistently in local environment
- [ ] Re-enable tests in CI workflow configuration
- [ ] Document any important findings for future reference

### DAB-Related Transactions to Test

The following transactions are part of the DAB feature and may have associated e2e tests:

- `NodeCreateTransaction`
- `NodeUpdateTransaction`
- `NodeDeleteTransaction`

---

## ✅ Acceptance Criteria

- [ ] Address the problem described above
- [ ] All DAB e2e tests pass consistently
- [ ] DAB tests are enabled in CI workflows
- [ ] No regressions to other SDK tests
- [ ] Follow existing project conventions
- [ ] Pass all CI checks

---

## 📋 Contribution Guide

To help your contribution go as smoothly as possible, we recommend following these steps:

- [ ] Wait for assignment

Read [Workflow Guide](docs/training/workflow.md) for step-by-step workflow guidance.  
Read [README.md](README.md) for setup instructions.

See the [Signing Guide](docs/training/signing.md).

---

### References

- **[HIP-869: Dynamic Address Book - Stage 1](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-869.md)** - Introduces the AddressBookService and node management transactions
- **[HIP-1299: Node Account ID Refinements](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-1299.md)** - Refinements to account ID management in DAB

### Key Concepts

| Concept | Description |
|---------|-------------|
| **AddressBookService** | HAPI service for managing network nodes |
| **Node ID** | Immutable unique identifier for a consensus node |
| **Admin Key** | Key controlled by node operator for signing node transactions |
| **Solo-action** | Local testing tool integrated with SDK CI workflows |

### DAB Transaction Summary

```
NodeCreateTransaction:
  - account_id (required)
  - description (optional, max 100 bytes UTF-8)
  - gossip_endpoint (required, max 10 entries)
  - service_endpoint (required, max 8 entries)
  - gossip_ca_certificate (required)
  - grpc_certificate_hash (optional)
  - admin_key (required)

NodeUpdateTransaction:
  - node_id (required)
  - account_id, description, endpoints, certificates, admin_key (all optional)

NodeDeleteTransaction:
  - node_id (required)
```

If you have questions, the community is happy to help:

https://discord.com/channels/905194001349627914/1337424839761465364

Omswastik-11:
Closes #1004

### DIFF
--- a/src/sdk/tests/integration/NodeCreateTransactionIntegrationTests.cc
+++ b/src/sdk/tests/integration/NodeCreateTransactionIntegrationTests.cc
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
+#include "AccountCreateTransaction.h"
 #include "BaseIntegrationTest.h"
 #include "ED25519PrivateKey.h"
 #include "NodeCreateTransaction.h"
@@ -13,7 +14,6 @@ using namespace Hiero;
 class NodeCreateTransactionIntegrationTests : public BaseIntegrationTest
 {
 protected:
-  [[nodiscard]] const AccountId& getAccountId() const { return mAccountId; }
   [[nodiscard]] const std::vector<Endpoint>& getGossipEndpoints() const { return mGossipEndpoints; }
   [[nodiscard]] const std::vector<Endpoint>& getGrpcServiceEndpoints() const { return mGrpcServiceEndpoints; }
   [[nodiscard]] const std::vector<std::byte> getGossipCertificate() const
@@ -22,7 +22,6 @@ protected:
   }
 
 private:
-  const AccountId mAccountId = AccountId::fromString("0.0.4");
   const Endpoint endpoint1 = Endpoint().setDomainName("test.com").setPort(123);
   const Endpoint endpoint2 = Endpoint().setDomainName("test2.com").setPort(123);
   const std::vector<Endpoint> mGossipEndpoints = { endpoint1, endpoint2 };
@@ -60,11 +59,22 @@ TEST_F(NodeCreateTransactionIntegrationTests, CanExecuteNodeCreateTransaction)
   // Given
   const std::shared_ptr<PrivateKey> adminKey = ED25519PrivateKey::generatePrivateKey();
 
+  TransactionResponse accountCreateTxResponse;
+  ASSERT_NO_THROW(accountCreateTxResponse = AccountCreateTransaction()
+                                              .setKey(adminKey->getPublicKey())
+                                              .setInitialBalance(Hbar(10LL))
+                                              .execute(getTestClient()));
+
+  TransactionReceipt accountCreateTxReceipt;
+  ASSERT_NO_THROW(accountCreateTxReceipt = accountCreateTxResponse.getReceipt(getTestClient()));
+  ASSERT_TRUE(accountCreateTxReceipt.mAccountId.has_value());
+  const AccountId newAccountId = accountCreateTxReceipt.mAccountId.value();
+
   // When / Then
   TransactionResponse txResponse;
 
   ASSERT_NO_THROW(txResponse = NodeCreateTransaction()
-                                 .setAccountId(getAccountId())
+                                 .setAccountId(newAccountId)
                                  .setGossipEndpoints(getGossipEndpoints())
                                  .setServiceEndpoints(getGrpcServiceEndpoints())
                                  .setGossipCaCertificate(getGossipCertificate())
--- a/src/sdk/tests/integration/NodeCreateTransactionIntegrationTests.cc
+++ b/src/sdk/tests/integration/NodeCreateTransactionIntegrationTests.cc
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 #include "BaseIntegrationTest.h"
 #include "ED25519PrivateKey.h"
-#include "FileId.h"
-#include "FreezeTransaction.h"
 #include "NodeCreateTransaction.h"
 #include "TransactionRecord.h"
 #include "TransactionResponse.h"
@@ -16,24 +14,15 @@ class NodeCreateTransactionIntegrationTests : public BaseIntegrationTest
 {
 protected:
   [[nodiscard]] const AccountId& getAccountId() const { return mAccountId; }
-  [[nodiscard]] const FileId& getFileId() const { return mFileId; }
   [[nodiscard]] const std::vector<Endpoint>& getGossipEndpoints() const { return mGossipEndpoints; }
   [[nodiscard]] const std::vector<Endpoint>& getGrpcServiceEndpoints() const { return mGrpcServiceEndpoints; }
   [[nodiscard]] const std::vector<std::byte> getGossipCertificate() const
   {
     return internal::HexConverter::hexToBytes(mGossipCertificateDer);
   }
-  [[nodiscard]] const std::vector<std::byte> getFileHash() const
-  {
-    return internal::HexConverter::hexToBytes(mFileHash);
-  }
 
 private:
   const AccountId mAccountId = AccountId::fromString("0.0.4");
-  const FileId mFileId = FileId::fromString("0.0.150");
-  // The file hash needs to be taken from the network context to be correct
-  const std::string mFileHash =
-    "ce52a3c62cf51f046ae2f85ff1c895da2b32876d6aa74d2454b6de9d11f58344e5065c807af5f2a1eb5850b26d016c3f";
   const Endpoint endpoint1 = Endpoint().setDomainName("test.com").setPort(123);
   const Endpoint endpoint2 = Endpoint().setDomainName("test2.com").setPort(123);
   const std::vector<Endpoint> mGossipEndpoints = { endpoint1, endpoint2 };
@@ -82,21 +71,6 @@ TEST_F(NodeCreateTransactionIntegrationTests, CanExecuteNodeCreateTransaction)
                                  .setAdminKey(adminKey->getPublicKey())
                                  .freezeWith(&getTestClient())
                                  .sign(adminKey)
-                                 .execute(getTestClient()));
-
-  ASSERT_NO_THROW(txResponse = FreezeTransaction()
-                                 .setFreezeType(FreezeType::PREPARE_UPGRADE)
-                                 .setFileId(getFileId())
-                                 .setFileHash(getFileHash())
-                                 .freezeWith(&getTestClient())
-                                 .execute(getTestClient()));
-
-  ASSERT_NO_THROW(txResponse = FreezeTransaction()
-                                 .setFreezeType(FreezeType::FREEZE_UPGRADE)
-                                 .setStartTime(std::chrono::system_clock::now() + std::chrono::seconds(5))
-                                 .setFileId(getFileId())
-                                 .setFileHash(getFileHash())
-                                 .freezeWith(&getTestClient())
                                  .execute(getTestClient()));
 
   TransactionReceipt txReceipt;
--- a/src/sdk/tests/integration/NodeCreateTransactionIntegrationTests.cc
+++ b/src/sdk/tests/integration/NodeCreateTransactionIntegrationTests.cc
@@ -66,7 +66,7 @@ private:
 };
 
 //-----
-TEST_F(NodeCreateTransactionIntegrationTests, DISABLED_CanExecuteNodeCreateTransaction)
+TEST_F(NodeCreateTransactionIntegrationTests, CanExecuteNodeCreateTransaction)
 {
   // Given
   const std::shared_ptr<PrivateKey> adminKey = ED25519PrivateKey::generatePrivateKey();
--- a/src/sdk/tests/integration/NodeDeleteTransactionIntegrationTests.cc
+++ b/src/sdk/tests/integration/NodeDeleteTransactionIntegrationTests.cc
@@ -30,7 +30,7 @@ private:
 };
 
 //-----
-TEST_F(NodeDeleteTransactionIntegrationTests, DISABLED_CanExecuteNodeDeleteTransaction)
+TEST_F(NodeDeleteTransactionIntegrationTests, CanExecuteNodeDeleteTransaction)
 {
   // Given / When / Then
   TransactionResponse txResponse;
--- a/src/sdk/tests/integration/NodeDeleteTransactionIntegrationTests.cc
+++ b/src/sdk/tests/integration/NodeDeleteTransactionIntegrationTests.cc
@@ -1,8 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
+#include "AccountCreateTransaction.h"
 #include "BaseIntegrationTest.h"
+#include "ED25519PrivateKey.h"
+#include "NodeCreateTransaction.h"
 #include "NodeDeleteTransaction.h"
 #include "TransactionRecord.h"
 #include "TransactionResponse.h"
+#include "impl/HexConverter.h"
 
 #include <gtest/gtest.h>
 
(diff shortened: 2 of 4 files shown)


src/sdk/tests/integration/NodeCreateTransactionIntegrationTests.cc
rwalworth: **issue (blocking):** These `FreezeTransaction` calls (`PREPARE_UPGRADE` and `FREEZE_UPGRADE`) should be removed. They aren't part of testing `NodeCreateTransaction` functionality - the test name is `CanExecuteNodeCreateTransaction`, and the only assertion needed is that the node create transaction executes successfully.

A `FREEZE_UPGRADE` also pauses the consensus network for up to 5 seconds, which will disrupt the other tests running concurrently in the `-j 6` parallel CI batch.

The test should end after the `NodeCreateTransaction` receipt is retrieved (line 85). Remove lines 87–103 entirely.

src/sdk/tests/integration/NodeDeleteTransactionIntegrationTests.cc
rwalworth: **issue (blocking):** `mNodeId = 2`, but Solo in `dualMode` provisions two nodes with IDs 0 and 1 (accounts 0.0.3 and 0.0.4). Node ID 2 doesn't exist, so this test will fail with an invalid-node error rather than actually testing `NodeDeleteTransaction`.

This test needs to target a node that exists in the test environment - either node 0 or node 1 - or be written to first create a node (via `NodeCreateTransaction`) and then delete the one it just created, cleaning up after itself.

rwalworth: **issue (blocking):** Same issue as in `NodeCreateTransactionIntegrationTests.cc` - the `FreezeTransaction` calls (`PREPARE_UPGRADE` and `FREEZE_UPGRADE`) should be removed. They aren't part of testing `NodeDeleteTransaction` functionality and they disrupt the parallel CI run.

The test should end after line 39 (the `NodeDeleteTransaction` executes). Remove lines 41–57 entirely.

src/sdk/tests/integration/NodeUpdateTransactionIntegrationTests.cc
rwalworth: **issue (blocking):** `mNodeId = 2` here as well, which has the same problem as in `NodeDeleteTransactionIntegrationTests.cc` - node ID 2 doesn't exist in the Solo `dualMode` environment (nodes 0 and 1 are the only ones provisioned). This test will fail with an invalid-node error.

The other enabled tests in this file use `nodeIDToUpdate = 1`. This test should target an existing node ID.

Additionally, once the node ID is corrected, this test modifies the node's account ID, endpoints, and admin key without resetting them afterward. The admin key is set to a freshly generated key that is not persisted, which would leave the node in an unrecoverable state for any subsequent operations. Cleanup logic (resetting the changed properties back to their original values) needs to be added at the end of the test, following the pattern used in `CanChangeNodeAccountUpdateAddressbookAndRetry`.

rwalworth (changes requested):
Thanks for the contribution @Omswastik-11! There's more to re-enabling these tests than just removing the `DISABLED_` prefix from the test names - the test code itself needs to be corrected first. I've left comments on the files below.

For reference, the [Hiero Swift SDK](https://github.com/hiero-ledger/hiero-sdk-swift) has a more complete set of DAB integration tests that are already enabled in CI and can serve as a guide for how DAB transactions and Solo behave - particularly `NodeCreateTransactionIntegrationTests.swift` and `NodeUpdateTransactionIntegrationTests.swift`. You're not required to add all the test cases that SDK has here (that can be a follow-up issue), but if you have questions about how DAB and Solo interact, you could look there.

src/sdk/tests/integration/NodeCreateTransactionIntegrationTests.cc
```diff
@@ -84,21 +73,6 @@ TEST_F(NodeCreateTransactionIntegrationTests, DISABLED_CanExecuteNodeCreateTrans
                                  .sign(adminKey)
                                  .execute(getTestClient()));
```
rwalworth, line 84: **issue (blocking):** CI is failing here with `ACCOUNT_IS_LINKED_TO_A_NODE`. Account `0.0.4` is already the account for node 1, so the network rejects using it as the account for a new node.

The fix is to create a fresh account inside the test body and pass that to `NodeCreateTransaction` instead - which is the same pattern the [Hiero Swift SDK uses](https://github.com/hiero-ledger/hiero-sdk-swift/blob/main/Tests/HieroIntegrationTests/NodeCreateTransactionIntegrationTests.swift). You'll also be able to remove the `mAccountId` field and `getAccountId()` accessor from the class fixture once they're no longer needed.

src/sdk/tests/integration/NodeDeleteTransactionIntegrationTests.cc
```diff
@@ -15,44 +12,20 @@ class NodeDeleteTransactionIntegrationTests : public BaseIntegrationTest
...
 
 //-----
-TEST_F(NodeDeleteTransactionIntegrationTests, DISABLED_CanExecuteNodeDeleteTransaction)
+TEST_F(NodeDeleteTransactionIntegrationTests, CanExecuteNodeDeleteTransaction)
 {
   // Given / When / Then
   TransactionResponse txResponse;
 
   ASSERT_NO_THROW(
     txResponse = NodeDeleteTransaction().setNodeId(getNodeId()).freezeWith(&getTestClient()).execute(getTestClient()));
```
rwalworth: **question:** This test deletes node ID 1 from the Solo network and expects a successful receipt, but it doesn't restore node 1 afterward. The `NodeUpdateTransactionIntegrationTests` run separately in CI (via `-R NodeUpdateTransactionIntegrationTests` after the main ctest suite) and most of those tests target node ID 1 via `nodeIDToUpdate`.

If deleting node 1 takes immediate effect in the Solo environment, those update tests could fail. Could you confirm:

1. That the test actually succeeds in Solo's `dualMode` (i.e., Solo allows deleting a genesis node via `NodeDeleteTransaction`)
2. That if it does succeed, the deletion doesn't break `NodeUpdateTransactionIntegrationTests` - for example, because DAB changes in Solo don't take effect until a network restart

If there's a risk of test interference, the safer pattern would be to first create a new node (getting its returned ID from the receipt), then delete that freshly created node instead.

src/sdk/tests/integration/NodeUpdateTransactionIntegrationTests.cc
```diff
@@ -25,97 +24,44 @@ class NodeUpdateTransactionIntegrationTests : public BaseIntegrationTest
...
-    "33bb07b97bf879b64487b447df510897d2a0a4e789c409a9b237a6ad240ad5464f2ce80c58ddc4d07a29a74eb25e1223db6c00e334d7a27d32"
-    "bfa6183a82f5e35bccf497c2445a526eabb0c068aba9b94cc092ea4756b0dcfb574f6179f0089e52b174ccdbd04123eeb6d70daeabd8513fcb"
-    "a6be0bc2b45ca9a69802dae11cc4d9ff6053b3a87fd8b0c6bf72fffc3b81167f73cca2b3fd656c5d353c8defca8a76e2ad535f984870a590af"
-    "4e28fed5c5a125bf360747c5e7742e7813d1bd39b5498c8eb6ba72f267eda034314fdbc596f6b967a0ef8be5231d364e634444c84e64bd7919"
-    "425171016fcd9bb05f01c58a303dee28241f6e860fc3aac3d92aad7dac2801ce79a3b41a0e1f1509fc0d86e96d94edb18616c000152490f645"
-    "61713102128990fedd3a5fa642f2ff22dc11bc4dc5b209986a0c3e4eb2bdfdd40e9fdf246f702441cac058dd8d0d51eb0796e2bea2ce1b37b2"
-    "a2f468505e1f8980a9f66d719df034a6fbbd2f9585991d259678fb9a4aebdc465d22c240351ed44abffbdd11b79a706fdf7c40158d3da87f68"
-    "d7bd557191a8016b5b899c07bf1b87590feb4fa4203feea9a2a7a73ec224813a12b7a21e5dc93fcde4f0a7620f570d31fe27e9b8d65b74db7d"
-    "c18a5e51adc42d7805d4661938";
+  const uint64_t mNodeId = 1;
```
rwalworth: **nitpick:** `mNodeId` (returned by `getNodeId()`) and `nodeIDToUpdate` now both equal `1` and represent the same thing. Minor redundancy, but not worth a separate cleanup PR - just something to tidy up if you're in this area again.

rwalworth (changes requested):
Thanks for working through those issues @Omswastik-11 - the new `CanExecuteNodeUpdateTransaction` implementation is well-done and querying the address book for the current description, applying a change, and reverting cleanly is exactly the right pattern.

Unfortunately `CanExecuteNodeCreateTransaction` is failing in CI. I left a comment below with what's going wrong and how to fix it.

src/sdk/tests/integration/NodeDeleteTransactionIntegrationTests.cc
Omswastik-11: Thanks for suggestion
I addressed this by changing `NodeDeleteTransactionIntegrationTests.cc` to create a fresh node first and then delete that newly created node, instead of deleting node 1. That keeps the test isolated and avoids any interference with `NodeUpdateTransactionIntegrationTests.cc` , which still targets node 1. So I’m not relying on Solo supporting deletion of a genesis node anymore.

src/sdk/tests/integration/NodeUpdateTransactionIntegrationTests.cc
```diff
@@ -25,97 +24,38 @@ class NodeUpdateTransactionIntegrationTests : public BaseIntegrationTest
 {
 protected:
   [[nodiscard]] const uint64_t& getNodeId() const { return mNodeId; }
```
rwalworth, line 26: **issue (blocking):** `getNodeId()` references `mNodeId`, but the `private:` section that defined it (`const uint64_t mNodeId = 1;`) was removed. This will cause a compile error.

The cleanup intention was right - both `mNodeId` and `nodeIDToUpdate` were `1` and meant the same thing, so removing the redundancy makes sense. But `getNodeId()` still needs a backing field. Please add back just the `mNodeId` declaration:

```cpp
private:
  const uint64_t mNodeId = 1;
```

This goes after `getNodeId()` (line 26) and before the closing `};`. The `nodeIDToUpdate` and all the endpoint/cert fields that were removed are fine to leave out - just `mNodeId` needs to come back.

rwalworth (changes requested):
Thanks for working through all the NodeCreate and NodeDelete feedback @Omswastik-11 - those two tests look exactly right now. The create-then-delete pattern in `NodeDeleteTransactionIntegrationTests.cc` is clean, and the account creation fix in `NodeCreateTransactionIntegrationTests.cc` directly addresses the `ACCOUNT_IS_LINKED_TO_A_NODE` failure.

There's one more issue I caught in `NodeUpdateTransactionIntegrationTests.cc`. One small fix needed below and we should be good to go.

Omswastik-11:
Hi @rwalworth !! Let me know if further changes are required . I have sent you my proposal and asked you for your suggestion in Discord DM . Do check that when you are free . Thanks for your guidance so far 🫡

rwalworth (approved):
LGTM - all three tests look correct now. The `mNodeId` private field is back, `CanExecuteNodeUpdateTransaction` cleanly reads the current description, applies the change, and reverts it, and `CanExecuteNodeDeleteTransaction` follows the create-then-delete pattern that avoids any interference with the other tests. Running the workflows and will merge once they pass!

I know this one took a few rounds to get through, @Omswastik-11 - thanks for sticking with the feedback and getting it right. Appreciate the patience.

rwalworth (changes requested):
The failing test is `CanExecuteNodeUpdateTransaction`. It gets back `INVALID_NODE_ACCOUNT_ID` from the network. The root cause is that the test uses `getTestClient()`, which submits transactions to node `0.0.3` at `127.0.0.1:50211` (from `local_node.json`). In the CI Solo environment, submitting a `NodeUpdateTransaction` for DAB node 1 (account `0.0.4`) through node `0.0.3` passes precheck but fails at consensus with `INVALID_NODE_ACCOUNT_ID`.

Every other passing `NodeUpdateTransaction` test sets up its own client targeting `0.0.4` at `localhost:51211` instead:

```cpp
std::unordered_map<std::string, AccountId> network;
network["localhost:51211"] = AccountId(4ULL);
Client client = Client::forNetwork(network);
client.setMirrorNetwork({ "localhost:5600" });
const std::string operatorKeyStr =
  "302e020100300506032b65700422042091132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137";
client.setOperator(AccountId(2ULL), ED25519PrivateKey::fromString(operatorKeyStr));
```

Please update `CanExecuteNodeUpdateTransaction` to follow that same pattern - replace every `getTestClient()` call with a locally configured `client` built as above, and pass that `client` to the `AddressBookQuery`, the update transaction, and the cleanup transaction. Once that's fixed the build should be green!

src/sdk/tests/integration/NodeUpdateTransactionIntegrationTests.cc
```diff
@@ -25,97 +24,52 @@ class NodeUpdateTransactionIntegrationTests : public BaseIntegrationTest
...
-    "61713102128990fedd3a5fa642f2ff22dc11bc4dc5b209986a0c3e4eb2bdfdd40e9fdf246f702441cac058dd8d0d51eb0796e2bea2ce1b37b2"
-    "a2f468505e1f8980a9f66d719df034a6fbbd2f9585991d259678fb9a4aebdc465d22c240351ed44abffbdd11b79a706fdf7c40158d3da87f68"
-    "d7bd557191a8016b5b899c07bf1b87590feb4fa4203feea9a2a7a73ec224813a12b7a21e5dc93fcde4f0a7620f570d31fe27e9b8d65b74db7d"
-    "c18a5e51adc42d7805d4661938";
+  const uint64_t mNodeId = 1;
 };
 
 //-----
-TEST_F(NodeUpdateTransactionIntegrationTests, DISABLED_CanExecuteNodeUpdateTransaction)
+TEST_F(NodeUpdateTransactionIntegrationTests, CanExecuteNodeUpdateTransaction)
```
rwalworth, line 33: **issue (blocking):** The test is still failing with `INVALID_NODE_ACCOUNT_ID`. The local client is now set up, but it's pointing to node 1's endpoint (`localhost:51211`, account `0.0.4`) and targeting node 1 via `getNodeId()`. Solo doesn't allow modifying node 1's description this way.

Both the [Java SDK](https://github.com/hiero-ledger/hiero-sdk-java) and [Swift SDK](https://github.com/hiero-ledger/hiero-sdk-swift) reference implementations for this test connect to **node 0** (`localhost:50211`, account `0.0.3`) and update **node 0**. Please bring `CanExecuteNodeUpdateTransaction` in line with that pattern:

```cpp
std::unordered_map<std::string, AccountId> network;
network["localhost:50211"] = AccountId(3ULL);
Client client = Client::forNetwork(network);
client.setMirrorNetwork({ "localhost:5600" });
client.setOperator(AccountId(2ULL), ED25519PrivateKey::fromString(
    "302e020100300506032b65700422042091132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137"));
```

Also replace `getNodeId()` with `0ULL` everywhere inside this test - in the `AddressBookQuery` `find_if` predicate, both `NodeUpdateTransaction` calls, and the cleanup call. The account-ID-change tests that currently use `getNodeId()` are correctly targeting node 1 and don't need to change.

src/sdk/tests/integration/NodeUpdateTransactionIntegrationTests.cc
```diff
@@ -25,97 +24,52 @@ class NodeUpdateTransactionIntegrationTests : public BaseIntegrationTest
...
-  ASSERT_NO_THROW(txResponse = FreezeTransaction()
-                                 .setFreezeType(FreezeType::FREEZE_UPGRADE)
-                                 .setStartTime(std::chrono::system_clock::now() + std::chrono::seconds(5))
-                                 .setFileId(getFileId())
-                                 .setFileHash(getFileHash())
-                                 .freezeWith(&getTestClient())
-                                 .execute(getTestClient()));
+  ASSERT_NO_THROW(
+    txResponse =
+      NodeUpdateTransaction().setNodeId(getNodeId()).setDescription(updatedDescription).execute(client));
```
rwalworth: **chore (blocking):** The clang-format check is still failing on the `ASSERT_NO_THROW(txResponse = ...)` block - the indentation doesn't match what clang-format-17 expects. After fixing the node 0 issue above, run clang-format on the file before pushing:

```bash
clang-format-17 -i src/sdk/tests/integration/NodeUpdateTransactionIntegrationTests.cc
```

If you don't have clang-format-17 installed, let me know and I can help you get it set up.

rwalworth (changes requested):
Thanks for adding the local client setup - that was the right direction, but CI is still failing on `CanExecuteNodeUpdateTransaction`. There are two issues left in that test, detailed below.

All 8 other `NodeUpdateTransactionIntegrationTests` pass, and `NodeCreateTransactionIntegrationTests.CanExecuteNodeCreateTransaction` and `NodeDeleteTransactionIntegrationTests.CanExecuteNodeDeleteTransaction` both pass as well.

src/sdk/tests/integration/NodeUpdateTransactionIntegrationTests.cc
Omswastik-11: I have just forgot to lint in last commit.

rwalworth (changes requested):
Good progress - lint is now clean, `NodeCreateTransactionIntegrationTests` and `NodeDeleteTransactionIntegrationTests` both pass, and the `CanExecuteNodeUpdateTransaction` test setup is structurally correct (node 0 client, `0ULL` everywhere). The remaining failure isn't in the test code - it's a bug in the SDK itself.

`NodeUpdateTransaction.cc` line 206 (`body->set_allocated_account_id(mAccountId.toProtobuf().release())`) always serializes `mAccountId` regardless of whether `setAccountId()` was ever called. A default-constructed `AccountId` serializes as `0.0.0`, which doesn't exist as a node account - so any `NodeUpdateTransaction` that doesn't explicitly call `setAccountId()` will fail at consensus with `INVALID_NODE_ACCOUNT_ID`.

The fix is to gate the two unconditional uses of `mAccountId` on a non-default check. `AccountId()` default-constructs to `0.0.0`, which is never a valid node account, so `!(mAccountId == AccountId())` cleanly distinguishes "not set" from "explicitly set". No new fields, no header change, no unit test updates.

**`src/sdk/main/src/NodeUpdateTransaction.cc`** - two spots:

```cpp
// build() - gate the serialization:
if (!(mAccountId == AccountId()))
{
  body->set_allocated_account_id(mAccountId.toProtobuf().release());
}

// validateChecksums() - gate the validation:
if (!(mAccountId == AccountId()))
{
  mAccountId.validateChecksum(client);
}
```

`initFromSourceTransactionBody()` already correctly guards with `body.has_account_id()` - no change needed there. No changes needed to `NodeUpdateTransactionUnitTests.cc`.

rwalworth (approved):
All tests pass. The SDK bug (`mAccountId` unconditionally serializing as `0.0.0` when `setAccountId()` was never called) is fixed, the test setup correctly targets node 0, and lint is clean. Good work seeing this through @Omswastik-11! Merging now

PR #1533, opened 2026-05-02
fix: resolve network update thread deadlock by releasing mutex before…

### LINKED ISSUE #931: [Advanced]: Investigate and Fix Client Network Update Thread Deadlock
## 🧠 Advanced

This issue is well-suited for contributors who are very familiar with the Hiero C++ SDK and enjoy working with its core abstractions and design patterns.

Advanced Issues often involve:
- Exploring and shaping SDK architecture  
- Reasoning about trade-offs and long-term impact  
- Working across multiple modules or systems  
- Updating tests, examples, and documentation alongside code  

The goal is to support thoughtful, high-impact contributions in a clear and collaborative way.

---

## 🐞 Problem Description

The Client's network update thread functionality is **currently disabled** due to a deadlock issue. The code that starts the thread is commented out in `Client::startNetworkUpdateThread()`:

```cpp
// src/sdk/main/src/Client.cc, line 1127-1128
void Client::startNetworkUpdateThread(const std::chrono::system_clock::duration& period)
{
  mImpl->mStartNetworkUpdateWaitTime = std::chrono::system_clock::now();
  mImpl->mNetworkUpdatePeriod = period;
  // mImpl->mNetworkUpdateThread =
  // std::make_unique<std::thread>(&Client::scheduleNetworkUpdate, this);
}
```

### Current Behavior

When the network update thread is enabled, **deadlocks occur** during:
- Integration test execution
- Other scenarios involving concurrent Client operations

### Root Cause (Suspected)

The deadlock appears to originate from lock contention. The network update thread:

1. Acquires `mImpl->mMutex` in `scheduleNetworkUpdate()`:
```cpp
if (std::unique_lock lock(mImpl->mMutex); !mImpl->mConditionVariable.wait_for(...))
{
  // While holding the lock, executes a query
  setNetworkFromAddressBookInternal(AddressBookQuery().setFileId(FileId::ADDRESS_BOOK).execute(*this));
}
```

2. While holding the lock, calls `execute(*this)` which likely attempts to acquire the same or a conflicting lock, causing deadlock.

### Impact

Without the network update thread:
- The SDK cannot automatically refresh network node information
- Users must manually update the network if nodes change
- The `DEFAULT_NETWORK_UPDATE_INITIAL_DELAY` and `DEFAULT_NETWORK_UPDATE_PERIOD` configurations have no effect

---

## 💡 Proposed / Expected Outcome

Investigate and fix the deadlock so the network update thread can be safely re-enabled.

### Goals

1. **Identify the exact deadlock**: Determine which locks are involved and where the circular dependency occurs
2. **Fix the synchronization**: Redesign the locking strategy to avoid deadlock while maintaining thread safety
3. **Re-enable the thread**: Uncomment and test the network update thread
4. **Verify stability**: Ensure integration tests pass reliably with the thread enabled

### Constraints

- Must maintain thread safety for concurrent Client operations
- Should not introduce performance regressions
- Must be backwards compatible (no public API changes)
- Should handle edge cases (Client move, destruction during update, etc.)

---

### Possible Approaches

1. **Release lock before query execution**: Release `mMutex` before calling `execute()`, then re-acquire to update state
   - Risk: State may change between release and re-acquire

2. **Use separate locks**: Introduce fine-grained locking for network updates vs. other Client operations
   - Risk: Increased complexity, potential for other deadlocks

3. **Lock-free signaling**: Use atomic flags and condition variables without holding locks during long operations
   - Risk: More complex state management

4. **Async/callback pattern**: Execute the query asynchronously and update state in a callback
   - Risk: Changes execution model

### Key Files to Investigate

| File | Purpose |
|------|---------|
| `src/sdk/main/src/Client.cc` | Network update thread, mutex, condition variable |
| `src/sdk/main/include/Client.h` | ClientImpl struct with synchronization primitives |
| `src/sdk/main/src/Executable.cc` | Query execution that may acquire locks |
| `src/sdk/main/src/impl/Network.cc` | Network operations that may be involved |

### Debugging Suggestions

1. **Enable thread sanitizer**: Build with `-fsanitize=thread` to detect data races
2. **Add logging**: Instrument lock acquisitions to trace the deadlock
3. **Minimal reproduction**: Create a focused test case that reliably triggers the deadlock
4. **Lock hierarchy analysis**: Map out all locks and their acquisition order

### Synchronization Primitives Currently Used

```cpp
// In ClientImpl (Client.cc)
std::mutex mMutex;                           // Main mutex
std::condition_variable mConditionVariable;  // For network update signaling
bool mCancelUpdate = false;                  // Cancellation flag
std::unique_ptr<std::thread> mNetworkUpdateThread;
```

---

## ✅ Acceptance Criteria

- [ ] **Identify the deadlock**: Document where and why the deadlock occurs
- [ ] **Fix the synchronization issue**: Implement a solution that avoids deadlock
- [ ] **Re-enable the network update thread**: Uncomment the thread creation code
- [ ] **Pass integration tests**: All existing tests pass with the thread enabled
- [ ] **Add regression test**: Create a test that would have caught this deadlock
- [ ] **Document the fix**: Explain the solution in code comments and/or PR description
- [ ] **No performance regression**: Thread operations don't significantly impact performance
- [ ] **Pass all CI checks**

---

### Related Issues/PRs

| Link | Description |
|------|-------------|
| [PR #930](https://github.com/hiero-ledger/hiero-sdk-cpp/pull/930) | Where this issue was discovered |

### Reproduction Steps

1. Uncomment lines 1127-1128 in `src/sdk/main/src/Client.cc`
2. Build the SDK
3. Run integration tests
4. Observe deadlock/hang

### Operating System

Reported on macOS, but likely affects all platforms.

### Version

Affects all versions (the thread has been disabled since its introduction).

---

If you have questions while working on this issue, feel free to ask!

You can reach the community and maintainers here:
[Hiero-SDK-C++ Discord](https://discord.com/channels/905194001349627914/1337424839761465364)

Whether you need help finding the right file, understanding existing code, or confirming your approach — we're happy to help.

aashu2006:
**Description**:

Fix network update thread deadlock by releasing mutex before query execution

* Release `mMutex` before calling `AddressBookQuery::execute()` in `scheduleNetworkUpdate()` to prevent self-deadlock on the same thread (non-recursive mutex)
* Fix secondary deadlock in `close()` and `setNetworkUpdatePeriod()` - call `cancelScheduledNetworkUpdate()` before acquiring `mMutex`, since `join()` waits for the bg thread which needs the same mutex
* Make `cancelScheduledNetworkUpdate()` hold `mMutex` briefly when setting `mCancelUpdate` and notifying, then release before `join()`
* Add null-safety guard for mirror/consensus network in the update loop to prevent nullptr dereference
* Re-enable the network update thread in `startNetworkUpdateThread()`
* Add 5 regression tests covering destruction, concurrent getters, period change, move, and close

Fixes #931

**Notes for reviewer**:

The root cause was a self-deadlock: `scheduleNetworkUpdate()` held `mImpl->mMutex` and called `AddressBookQuery::execute(*this)`, which calls `client.getRequestTimeout()` (and many other getters) - all of which try to re-lock the same non-recursive `std::mutex` on the same thread.

The fix follows the same pattern already used in `updateAddressBook()` (line 703): execute the query without the lock, then re-acquire to update state. A cancellation check after re-acquiring handles the race where someone calls `close()` during the query window.

**Checklist**

### DIFF
--- a/src/sdk/main/src/Client.cc
+++ b/src/sdk/main/src/Client.cc
@@ -588,8 +588,12 @@ std::optional<std::function<std::vector<std::byte>(const std::vector<std::byte>&
 
 void Client::close()
 {
-  std::unique_lock lock(mImpl->mMutex);
+  // Cancel the network update thread first WITHOUT holding the mutex.
+  // cancelScheduledNetworkUpdate() calls join() on the background thread,
+  // and that thread needs mMutex — holding the mutex here would deadlock.
   cancelScheduledNetworkUpdate();
+
+  std::unique_lock lock(mImpl->mMutex);
 
   std::for_each(mImpl->mSubscriptions.begin(),
                 mImpl->mSubscriptions.end(),
@@ -771,10 +775,12 @@ Logger Client::getLogger() const
 //-----
 Client& Client::setNetworkUpdatePeriod(const std::chrono::system_clock::duration& update)
 {
-  std::unique_lock lock(mImpl->mMutex);
-
-  // Cancel any previous network updates and wait for the thread to complete.
+  // Cancel any previous network updates WITHOUT holding the mutex.
+  // cancelScheduledNetworkUpdate() calls join() on the background thread,
+  // and that thread needs mMutex — holding the mutex here would deadlock.
   cancelScheduledNetworkUpdate();
+
+  std::unique_lock lock(mImpl->mMutex);
 
   // Update the network update period.
   mImpl->mNetworkUpdatePeriod = update;
@@ -1144,8 +1150,7 @@ void Client::startNetworkUpdateThread(const std::chrono::system_clock::duration&
 {
   mImpl->mStartNetworkUpdateWaitTime = std::chrono::system_clock::now();
   mImpl->mNetworkUpdatePeriod = period;
-  // mImpl->mNetworkUpdateThread =
-  // std::make_unique<std::thread>(&Client::scheduleNetworkUpdate, this);
+  mImpl->mNetworkUpdateThread = std::make_unique<std::thread>(&Client::scheduleNetworkUpdate, this);
 }
 
 //-----
@@ -1154,33 +1159,67 @@ void Client::scheduleNetworkUpdate()
   // Network updates should keep occurring until they're cancelled.
   while (true)
   {
-    if (std::unique_lock lock(mImpl->mMutex); !mImpl->mConditionVariable.wait_for(
-          lock, mImpl->mNetworkUpdatePeriod, [this]() { return mImpl->mCancelUpdate; }))
+    // Wait for the update period while holding the lock (condition_variable
+    // requires the lock). If the wait was NOT cancelled (returns false), we
+    // need to perform a network update.
+    bool shouldUpdate = false;
     {
-      try
-      {
-        // Get the address book and set the network based on the address book.
-        setNetworkFromAddressBookInternal(AddressBookQuery().setFileId(FileId::ADDRESS_BOOK).execute(*this));
+      std::unique_lock lock(mImpl->mMutex);
+      shouldUpdate = !mImpl->mConditionVariable.wait_for(
+        lock, mImpl->mNetworkUpdatePeriod, [this]() { return mImpl->mCancelUpdate; });
+    }
+    // Lock is released here.
 
-        // Adjust the network update period if this is the initial update.
-        if (!mImpl->mMadeInitialNetworkUpdate)
-        {
-          mImpl->mNetworkUpdatePeriod = DEFAULT_NETWORK_UPDATE_PERIOD;
-          mImpl->mMadeInitialNetworkUpdate = true;
-        }
+    if (!shouldUpdate)
+    {
+      // The network update was cancelled, stop looping.
+      break;
+    }
 
-        // Schedule the next network update.
-        mImpl->mStartNetworkUpdateWaitTime = std::chrono::system_clock::now();
-      }
-      catch (const std::exception& exception)
+    // Skip the update if no mirror network or consensus network is configured.
+    // AddressBookQuery requires a mirror network endpoint, and
+    // setNetworkFromAddressBookInternal requires a consensus network.
+    // Calling execute() without a mirror network would dereference a nullptr.
+    if (!getClientMirrorNetwork() || !getClientNetwork())
+    {
+      continue;
+    }
+
+    try
+    {
+      // Execute the address book query WITHOUT holding mMutex.
+      // AddressBookQuery::execute() calls back into Client getters
+      // (getRequestTimeout, getClientMirrorNetwork, etc.) which all
+      // acquire mMutex. Holding mMutex here would cause a self-deadlock
+      // on the same thread since std::mutex is non-recursive.
+      const NodeAddressBook addressBook = AddressBookQuery().setFileId(FileId::ADDRESS_BOOK).execute(*this);
+
+      // Re-acquire the lock to update internal state.
+      std::unique_lock lock(mImpl->mMutex);
+
+      // Check if we were cancelled while the query was executing.
+      if (mImpl->mCancelUpdate)
       {
-        mImpl->mLogger.warn(std::string("Failed to update address book via mirror node query ") + exception.what());
         break;
       }
+
+      // Update the network from the fetched address book.
+      setNetworkFromAddressBookInternal(addressBook);
+
+      // Adjust the network update period if this is the initial update.
+      if (!mImpl->mMadeInitialNetworkUpdate)
+      {
+        mImpl->mNetworkUpdatePeriod = DEFAULT_NETWORK_UPDATE_PERIOD;
+        mImpl->mMadeInitialNetworkUpdate = true;
+      }
+
+      // Schedule the next network update.
+      mImpl->mStartNetworkUpdateWaitTime = std::chrono::system_clock::now();
     }
-    // The network update was cancelled, stop looping.
-    else
+    catch (const std::exception& exception)
     {
+      std::unique_lock lock(mImpl->mMutex);
+      mImpl->mLogger.warn(std::string("Failed to update address book via mirror node query ") + exception.what());
       break;
     }
   }
@@ -1195,9 +1234,17 @@ void Client::cancelScheduledNetworkUpdate()
     return;
   }
 
-  // Signal the thread to stop and wait for it to stop (if it was running).
-  mImpl->mCancelUpdate = true;
-  mImpl->mConditionVariable.notify_all();
+  // Signal the thread to stop. Must hold the mutex when modifying
+  // mCancelUpdate and notifying, since the background thread checks
+  // the flag under the same mutex in its condition_variable predicate.
+  {
+    std::unique_lock lock(mImpl->mMutex);
+    mImpl->mCancelUpdate = true;
+    mImpl->mConditionVariable.notify_all();
+  }
+  // Lock released before join() — the background thread needs mMutex
+  // to complete its current wait_for or to re-acquire after query execution.
+
   if (mImpl->mNetworkUpdateThread->joinable())
   {
     mImpl->mNetworkUpdateThread->join();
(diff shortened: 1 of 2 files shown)


aashu2006:
@rwalworth Let me know if any tweaks are required for this or any suggestion. Thanks!

src/sdk/main/src/Client.cc
```diff
@@ -1154,33 +1159,67 @@ void Client::scheduleNetworkUpdate()
   // Network updates should keep occurring until they're cancelled.
   while (true)
   {
-    if (std::unique_lock lock(mImpl->mMutex); !mImpl->mConditionVariable.wait_for(
-          lock, mImpl->mNetworkUpdatePeriod, [this]() { return mImpl->mCancelUpdate; }))
+    // Wait for the update period while holding the lock (condition_variable
+    // requires the lock). If the wait was NOT cancelled (returns false), we
+    // need to perform a network update.
+    bool shouldUpdate = false;
```
rwalworth: **nitpick:** The variable naming and logic here are correct - `shouldUpdate = true` means the timeout fired and we should update, `if (!shouldUpdate) break` correctly exits on cancellation. No issue with the code itself.

The comment above it is slightly indirect though:

> "If the wait was NOT cancelled (returns false), we need to perform a network update."

A future reader might parse "returns false" as an odd way to describe the non-cancellation case. Something like this is a bit clearer:

```cpp
// wait_for returns true  if the predicate fired (cancelled) --> should NOT update
// wait_for returns false if the timeout expired             --> should update
// Negate so that shouldUpdate == true means "run the update".
```

Not worth holding up the PR over - just easier to read at a glance.

src/sdk/main/src/Client.cc
```diff
@@ -1154,33 +1159,67 @@ void Client::scheduleNetworkUpdate()
...
-      }
-      catch (const std::exception& exception)
+    // Skip the update if no mirror network or consensus network is configured.
+    // AddressBookQuery requires a mirror network endpoint, and
+    // setNetworkFromAddressBookInternal requires a consensus network.
+    // Calling execute() without a mirror network would dereference a nullptr.
+    if (!getClientMirrorNetwork() || !getClientNetwork())
+    {
+      continue;
+    }
```
rwalworth, line 1186: **thought:** The `continue` guard is a safe fix - it prevents the crash and the behavior is acceptable since factory methods (`forNetwork`, `forMainnet`, etc.) configure the network before the 10-second timer fires. Good incremental improvement.

One thing worth considering as a follow-up: the root reason this guard is needed is that the bg thread starts unconditionally at construction, before any network is configured. A more principled design would be to start the update cycle only once both a mirror network and a consensus network are set - i.e., triggered from `setNetwork()` / `setMirrorNetwork()` rather than from the constructor. That would:

- Eliminate the null-check spin entirely
- Make update timing relative to when the network was actually configured
- Avoid running a bg thread at all for clients that are never given a network

It's a bigger change (both setters would need to coordinate, `setNetworkUpdatePeriod` would need to handle the not-yet-started case, factory methods would need explicit thread starts), so it's not the right scope for this PR. But it would clean up the design and remove the need for this guard. Worth a follow-up issue if you're interested.

src/sdk/main/src/Client.cc
```diff
@@ -1154,33 +1159,67 @@ void Client::scheduleNetworkUpdate()
...
+      mImpl->mStartNetworkUpdateWaitTime = std::chrono::system_clock::now();
     }
-    // The network update was cancelled, stop looping.
-    else
+    catch (const std::exception& exception)
     {
+      std::unique_lock lock(mImpl->mMutex);
+      mImpl->mLogger.warn(std::string("Failed to update address book via mirror node query ") + exception.what());
       break;
     }
```
rwalworth, line 1224: **question:** The new exception handler acquires `mMutex` before calling `mImpl->mLogger.warn()`. This matches the old behavior (the old catch block was lexically inside the `if` that held the lock), so the thread-safety level is preserved - good.

One thing to confirm: is `mLogger.warn()` actually unsafe to call without the lock? If `mLogger` holds its own internal lock (which log4cxx typically does), then the outer lock on `mMutex` here is redundant. Not a correctness issue either way, just want to make sure we're not holding `mMutex` longer than necessary in a path that's already failing.

src/sdk/tests/unit/ClientUnitTests.cc
rwalworth, line 1: **suggestion (blocking, test):** The five regression tests all complete in well under 10 seconds, but `DEFAULT_NETWORK_UPDATE_INITIAL_DELAY` is 10 seconds - so the background thread never actually wakes up and runs in any of them. Every test exercises only the "cancel a sleeping thread" path (the secondary deadlock fix). The primary fix - releasing the lock before calling `execute()` - is not exercised by any test.

One path that *is* unit-testable right now without a mock network is the null-network `continue` branch added at the top of `scheduleNetworkUpdate`. Please add a test along these lines:

```cpp
//-----
TEST_F(ClientUnitTests, NetworkUpdateThreadSkipsUpdateWhenNoNetworkConfigured)
{
  // Given
  Client client;
  client.setNetworkUpdatePeriod(std::chrono::milliseconds(50));

  // When
  std::this_thread::sleep_for(std::chrono::milliseconds(300));

  // Then
  EXPECT_EQ(client.getNetworkUpdatePeriod(), std::chrono::milliseconds(50));
}
```

This is the only test that actually forces the thread past the `wait_for` and into the body of `scheduleNetworkUpdate`. The `getNetworkUpdatePeriod()` call at the end also doubles as a lightweight check that the mutex isn't being held by the bg thread (it would hang if it were).

rwalworth: **thought:** Even with my other comment addressed, the core deadlock scenario - a getter called while the bg thread is between "lock released" and "lock re-acquired after execute()" - still has no automated coverage. The `AddressBookQueryIntegrationTests` manually calls `execute()` directly and doesn't involve the bg thread. The integration base also actively suppresses the thread by calling `setNetworkUpdatePeriod(360s)`, so no integration test exercises it either.

Covering this properly would require either a mock `MirrorNetwork` that can be paused mid-response to create a controllable window, or a test-hook constructor that accepts a custom initial delay. That's a bigger lift and out of scope for this PR. I'd suggest opening a follow-up issue to track it - something like "Add unit test for concurrent getter access during bg network update" so it doesn't get lost.

rwalworth (changes requested):
Thanks for the deep work here @aashu2006! The root cause analysis is correct and the core fix is right - releasing the lock before calling `AddressBookQuery::execute()` matches the existing pattern in `updateAddressBook()` (line 703–715). The re-enablement of the thread is the right move now that the deadlock is resolved.

I went through every codepath the changes touch and left my findings below. One blocking ask (a missing test), one nitpick, and two design thoughts for follow-up.

src/sdk/tests/unit/ClientUnitTests.cc
aashu2006: yes this is out of scope here. A follow-up issue can be opened to track adding a proper concurrent access test with a mock `MirrorNetwork`.

src/sdk/main/src/Client.cc
aashu2006: Yes, starting the thread only once both networks are configured is the right long-term design. I can open a follow-up issue to track it.

src/sdk/main/src/Client.cc
aashu2006: Logger wraps `log4cxx::Logger` which is internally thread safe, so the outer `mMutex` is technically redundant here. I kept it to match the original code's pattern,  happy to remove it if you would prefer.

src/sdk/main/src/Client.cc
aashu2006: I've updated the comment to your suggested format now. Much clearer at a glance.

src/sdk/tests/unit/ClientUnitTests.cc
aashu2006: Also added, it uses a 50ms update period so the bg thread actually wakes up and hits the null-network continue guard. The `getNetworkUpdatePeriod()` call at the end also serves as a mutex-not-stuck check.

src/sdk/main/src/Client.cc
rwalworth: Sounds good, please do so. Thanks!

src/sdk/main/src/Client.cc
rwalworth: We don't need to hold up the merge here, we can open up a follow-up issue to remove it.

rwalworth (approved):
LGTM @aashu2006! Running the workflows now and will merge once they pass!

PR #1595, opened 2026-05-07
feat: add recovery ID calculation for Ethereum transaction signing

### LINKED ISSUE #684: [Advanced]: Add Recovery ID Calculation for Ethereum Transaction Signing
## 🧠 Advanced

This issue is well-suited for contributors who are very familiar with the Hiero C++ SDK and enjoy working with its core abstractions and design patterns.

Advanced Issues often involve:
- Exploring and shaping SDK architecture  
- Reasoning about trade-offs and long-term impact  
- Working across multiple modules or systems  
- Updating tests, examples, and documentation alongside code  

The goal is to support thoughtful, high-impact contributions in a clear and collaborative way.

---

## 🐞 Problem Description

When signing Ethereum transactions, the signature includes three components: `r`, `s`, and `v` (the recovery ID). The recovery ID is an index from 0 to 3 that indicates which of the 4 possible public keys can be recovered from the signature. This is required for proper Ethereum transaction construction and EVM compatibility.

Currently, the C++ SDK cannot calculate the recovery ID because **OpenSSL does not support recovery ID generation**. There is no built-in OpenSSL API for public key recovery from signatures - this is by design since OpenSSL focuses on TLS/PKI where recovery isn't needed.

The `EthereumTransactionDataEip1559` class has a `mRecoveryId` field, but there's no way to compute it.

This architectural limitation requires introducing a new cryptographic dependency (`secp256k1`) to the SDK.

Relevant areas:
- `src/sdk/main/include/ECDSAsecp256k1PrivateKey.h`
- `src/sdk/main/src/ECDSAsecp256k1PrivateKey.cc`
- `src/sdk/main/include/EthereumTransactionDataEip1559.h`
- `vcpkg.json` (dependency management)
- `vcpkg/ports/secp256k1/` (existing port with recovery enabled)

---

## 💡 Proposed / Expected Outcome

Add a `getRecoveryId` method to `ECDSAsecp256k1PrivateKey` using the `secp256k1` library (bitcoin-core) which supports recovery operations.

**Key design decisions:**

1. **Add secp256k1 as a dependency** - A vcpkg port already exists at `vcpkg/ports/secp256k1/` with `-DENABLE_MODULE_RECOVERY` enabled. Add it to `vcpkg.json`.

2. **Dual crypto library approach** - Use secp256k1 specifically for recovery operations while keeping OpenSSL for existing ECDSA operations. This minimizes disruption.

3. **New method signature:**

```cpp
/**
 * Computes the recovery ID for an ECDSA signature.
 *
 * @param r The r component of the ECDSA signature (32 bytes).
 * @param s The s component of the ECDSA signature (32 bytes).
 * @param message The original message that was signed (will be hashed with Keccak-256).
 * @return The recovery ID (0-3), or -1 if recovery fails.
 */
[[nodiscard]] int getRecoveryId(const std::vector<std::byte>& r,
                                 const std::vector<std::byte>& s,
                                 const std::vector<std::byte>& message) const;
```

**Constraints:**
- Preserve existing OpenSSL-based signing and verification
- Ensure cross-platform compatibility (secp256k1 builds on Linux, macOS, Windows)
- No breaking changes to public APIs

---

## 🧠 Implementation & Design Notes

**Why secp256k1 instead of OpenSSL:**

OpenSSL (as of 2025) does not support ECDSA recovery operations. This is confirmed and unlikely to change since it's outside OpenSSL's focus on TLS/PKI standards. The `secp256k1` library (bitcoin-core) is purpose-built for this and has a dedicated recovery module.

**Existing infrastructure:**

The SDK already has a vcpkg port prepared:

```cmake
# vcpkg/ports/secp256k1/CMakeLists.txt
add_definitions(
    -DENABLE_MODULE_RECOVERY    # Recovery support already enabled!
)
```

**Implementation approach:**

1. Add `secp256k1` to `vcpkg.json` dependencies
2. Update CMakeLists.txt to link the library
3. Implement `getRecoveryId` using:
   - `secp256k1_ecdsa_recoverable_signature_parse_compact()`
   - `secp256k1_ecdsa_recover()`
4. Hash message with Keccak-256 (consistent with existing ECDSA signing)
5. Try recovery IDs 0-3, return the one that matches expected public key

**Reference implementation (Swift SDK):**

The Swift SDK implements this in `PrivateKey.swift` lines 562-620 using the same approach with its secp256k1 wrapper.

**Testing strategy:**
- Unit tests with known Ethereum signatures
- Cross-platform build verification
- Integration tests with EthereumTransaction flows

---

## ✅ Acceptance Criteria

- [ ] Add `secp256k1` as a dependency in `vcpkg.json`
- [ ] Implement `getRecoveryId` method on `ECDSAsecp256k1PrivateKey`
- [ ] Method correctly computes recovery ID (0-3) for valid signatures
- [ ] Method returns -1 for invalid inputs or failed recovery
- [ ] Maintain backwards compatibility (no changes to existing APIs)
- [ ] Follow existing architectural and C++ conventions
- [ ] Include comprehensive unit tests
- [ ] Build succeeds on all supported platforms (Linux, macOS, Windows)
- [ ] Pass all CI checks

---

## 📚 Additional Context, Links, or Prior Art

**References:**
- [SEC1v2 Section 4.1.6](https://www.secg.org/sec1-v2.pdf) - ECDSA Public Key Recovery Operation
- [bitcoin-core/secp256k1](https://github.com/bitcoin-core/secp256k1) - Recovery module documentation

**Swift SDK implementation:**
- `hiero-sdk-swift/Sources/Hiero/PrivateKey.swift` lines 562-620
- `hiero-sdk-swift/Tests/HieroIntegrationTests/EthereumTransactionIntegrationTests.swift` lines 171-183

**Existing SDK infrastructure:**
- `vcpkg/ports/secp256k1/` - Port already prepared with recovery enabled
- `vcpkg/ports/secp256k1/CMakeLists.txt` - Shows `-DENABLE_MODULE_RECOVERY`

**Community support:**
https://discord.com/channels/905194001349627914/1337424839761465364

aashu2006:
**Description**:

Add recovery ID calculation for Ethereum transaction signing using bitcoin-core/secp256k1.

OpenSSL does not support ECDSA recovery operations, which are required for computing the `v` value in Ethereum transaction signatures. This introduces `secp256k1` as a targeted dependency specifically for recovery operations, while keeping OpenSSL for all existing ECDSA signing and verification.

* Add `secp256k1` as a vcpkg dependency
* Add `find_package(unofficial-secp256k1)` and link `unofficial::secp256k1` in CMake build system
* Implement `getRecoveryId(r, s, message)` method on `ECDSAsecp256k1PrivateKey`
* Add 6 unit tests covering valid recovery, non-deterministic signing consistency, empty message, invalid input sizes, and mismatched key detection

Fixes #684

**Notes for reviewer**:

The implementation uses a try-recover approach: for each candidate recovery ID (0–3), it recovers a public key using `secp256k1_ecdsa_recover()` and compares it against the expected public key. The message is hashed with Keccak-256, consistent with the existing `sign()` method.

Key design decisions:
- **Dual crypto library**: secp256k1 is used only for recovery; OpenSSL remains for all existing operations. No existing APIs are changed.
- **The vcpkg port already exists** at `vcpkg/ports/secp256k1/` with `-DENABLE_MODULE_RECOVERY` enabled - we just needed to add it to `vcpkg.json` and wire up CMake.
- **Method returns -1** on invalid inputs or failed recovery rather than throwing, matching the issue spec and keeping it lightweight for callers.

**Checklist**

### DIFF
--- a/src/sdk/main/src/ECDSAsecp256k1PrivateKey.cc
+++ b/src/sdk/main/src/ECDSAsecp256k1PrivateKey.cc
@@ -15,9 +15,15 @@
 #include "impl/openssl_utils/EVP_MD_CTX.h"
 #include "impl/openssl_utils/OSSL_LIB_CTX.h"
 #include "impl/openssl_utils/OpenSSLUtils.h"
+#include "impl/openssl_utils/Secp256k1Context.h"
 
 #include <openssl/ec.h>
 #include <openssl/x509.h>
+
+#include <secp256k1.h>
+#include <secp256k1_recovery.h>
+
+#include <cstring>
 
 #include <services/basic_types.pb.h>
 
@@ -69,6 +75,42 @@ const internal::OpenSSLUtils::BIGNUM CURVE_ORDER =
   }
 
   return key;
+}
+
+/**
+ * Attempt to recover a public key using the given recovery ID and compare it to the expected key.
+ *
+ * @param ctx              The secp256k1 context.
+ * @param compactSig       The compact signature (r || s) as 64 bytes.
+ * @param recId            The recovery ID to try (0-3).
+ * @param messageHash      The 32-byte Keccak-256 hash of the original message.
+ * @param expectedPubKey   The expected uncompressed public key bytes (65 bytes).
+ * @return True if the recovered public key matches the expected key.
+ */
+[[nodiscard]] bool tryRecoverWithId(const secp256k1_context* ctx,
+                                    const unsigned char* compactSig,
+                                    int recId,
+                                    const unsigned char* messageHash,
+                                    const std::vector<std::byte>& expectedPubKey)
+{
+  secp256k1_ecdsa_recoverable_signature recoverableSig;
+  if (secp256k1_ecdsa_recoverable_signature_parse_compact(ctx, &recoverableSig, compactSig, recId) != 1)
+  {
+    return false;
+  }
+
+  secp256k1_pubkey recoveredPubKey;
+  if (secp256k1_ecdsa_recover(ctx, &recoveredPubKey, &recoverableSig, messageHash) != 1)
+  {
+    return false;
+  }
+
+  // Serialize the recovered public key in uncompressed form.
+  unsigned char serializedPubKey[ECDSAsecp256k1PublicKey::UNCOMPRESSED_KEY_SIZE];
+  size_t outputLen = sizeof(serializedPubKey);
+  secp256k1_ec_pubkey_serialize(ctx, serializedPubKey, &outputLen, &recoveredPubKey, SECP256K1_EC_UNCOMPRESSED);
+
+  return outputLen == expectedPubKey.size() && std::memcmp(serializedPubKey, expectedPubKey.data(), outputLen) == 0;
 }
 
 } // namespace
@@ -261,6 +303,52 @@ std::vector<std::byte> ECDSAsecp256k1PrivateKey::sign(const std::vector<std::byt
   }
 
   return outputArray;
+}
+
+//-----
+int ECDSAsecp256k1PrivateKey::getRecoveryId(const std::vector<std::byte>& r,
+                                            const std::vector<std::byte>& s,
+                                            const std::vector<std::byte>& message) const
+{
+  // Validate input sizes.
+  if (r.size() != R_SIZE || s.size() != S_SIZE)
+  {
+    return -1;
+  }
+
+  // Compute the Keccak-256 hash of the message.
+  const std::vector<std::byte> messageHash = internal::OpenSSLUtils::computeKECCAK256(message);
+
+  // Build compact signature (r || s) as 64-byte array.
+  std::vector<unsigned char> compactSig(RAW_SIGNATURE_SIZE);
+  std::memcpy(compactSig.data(), r.data(), R_SIZE);
+  std::memcpy(compactSig.data() + R_SIZE, s.data(), S_SIZE);
+
+  // Get the expected uncompressed public key bytes (65 bytes: 0x04 || x || y).
+  const std::vector<std::byte> expectedPubKeyBytes =
+    ECDSAsecp256k1PublicKey::uncompressBytes(getPublicKey()->toBytesRaw());
+
+  // Create a secp256k1 context for verification/recovery using RAII.
+  internal::OpenSSLUtils::Secp256k1Context ctx(secp256k1_context_create(SECP256K1_CONTEXT_NONE));
+  if (!ctx)
+  {
+    return -1;
+  }
+
+  // Try each recovery ID (0-3) and check which one recovers our public key.
+  for (int recId = 0; recId < 4; ++recId)
+  {
+    if (tryRecoverWithId(ctx.get(),
+                         compactSig.data(),
+                         recId,
+                         reinterpret_cast<const unsigned char*>(messageHash.data()),
+                         expectedPubKeyBytes))
+    {
+      return recId;
+    }
+  }
+
+  return -1;
 }
 
 //-----
--- a/src/sdk/tests/unit/ECDSAsecp256k1PrivateKeyUnitTests.cc
+++ b/src/sdk/tests/unit/ECDSAsecp256k1PrivateKeyUnitTests.cc
@@ -4,6 +4,7 @@
 #include "ED25519PrivateKey.h"
 #include "exceptions/BadKeyException.h"
 #include "exceptions/UninitializedException.h"
+#include "impl/HexConverter.h"
 #include "impl/Utilities.h"
 
 #include <algorithm>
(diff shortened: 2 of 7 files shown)


src/sdk/tests/unit/ECDSAsecp256k1PrivateKeyUnitTests.cc
rwalworth, line 1: **issue (blocking, test):** None of the new tests verify that `getRecoveryId` returns the *correct* value - they only check that the result falls in `[0, 3]` or equals `-1`. With ECDSA being non-deterministic, you can't pin a recovery ID across calls - but you can pin one with a fixed signature input.

Please add at least one test with a known-good vector: a hardcoded `(privateKey, message, r, s)` tuple where the expected recovery ID is precisely known (computed once via the Java/Swift/Go SDK or a reference tool, then frozen here). Without this, an off-by-one in the recovery loop, an endianness bug in the compact-signature layout, or a wrong hash function could pass all six current tests silently.

A representative shape (illustrative - pick real values):

```cpp
TEST_F(ECDSAsecp256k1PrivateKeyUnitTests, GetRecoveryIdMatchesKnownVector)
{
  const std::unique_ptr<ECDSAsecp256k1PrivateKey> privateKey =
    ECDSAsecp256k1PrivateKey::fromString("<known hex>");
  const std::vector<std::byte> message = internal::HexConverter::hexToBytes("<known msg>");
  const std::vector<std::byte> r = internal::HexConverter::hexToBytes("<known r>");
  const std::vector<std::byte> s = internal::HexConverter::hexToBytes("<known s>");

  EXPECT_EQ(privateKey->getRecoveryId(r, s, message), <expected: 0 or 1>);
}
```

This is the kind of test that catches integration bugs we'd otherwise only find when the recovered transaction fails at the EVM layer.

src/sdk/tests/unit/ECDSAsecp256k1PrivateKeyUnitTests.cc
```diff
@@ -303,3 +303,116 @@ TEST_F(ECDSAsecp256k1PrivateKeyUnitTests, ECDSACompatibility)
...
+
+  // When
+  const std::vector<std::byte> signature = signingKey->sign(message);
+  const std::vector<std::byte> r(signature.cbegin(), signature.cbegin() + ECDSAsecp256k1PrivateKey::R_SIZE);
+  const std::vector<std::byte> s(signature.cbegin() + ECDSAsecp256k1PrivateKey::R_SIZE, signature.cend());
+  const int recoveryId = differentKey->getRecoveryId(r, s, message);
+
+  // Then - The recovery ID should not match the different key.
+  EXPECT_EQ(recoveryId, -1);
+}
```
rwalworth, line 441: **suggestion (non-blocking):** This test is logically correct, but the name "MismatchedSignature" reads a bit ambiguously. The semantics are really "calling `getRecoveryId` on a signature produced by a *different* key returns -1 because none of the four candidate recoveries match this key's pubkey."

Consider renaming to `GetRecoveryIdReturnsNegativeOneForForeignSignature` (or similar) and adding a one-line comment explaining the contract: "When a signature was produced by a different key, no recovery ID can recover this key's public key, so -1 is returned." Future readers will thank you - `EXPECT_EQ(recoveryId, -1)` for what looks like a valid signature is initially confusing.

src/sdk/main/src/ECDSAsecp256k1PrivateKey.cc
```diff
@@ -71,6 +76,43 @@ const internal::OpenSSLUtils::BIGNUM CURVE_ORDER =
...
+  }
+
+  secp256k1_pubkey recoveredPubKey;
+  if (secp256k1_ecdsa_recover(ctx, &recoveredPubKey, &recoverableSig, messageHash) != 1)
+  {
+    return false;
+  }
+
+  // Serialize the recovered public key in uncompressed form (65 bytes).
+  unsigned char serializedPubKey[65];
```
rwalworth: **suggestion (non-blocking):** There's already a named constant for this - [`ECDSAsecp256k1PublicKey::UNCOMPRESSED_KEY_SIZE`](src/sdk/main/include/ECDSAsecp256k1PublicKey.h#L28) (= 65). Please use that instead of the magic `65`:

```cpp
unsigned char serializedPubKey[ECDSAsecp256k1PublicKey::UNCOMPRESSED_KEY_SIZE];
size_t outputLen = sizeof(serializedPubKey);
```

Same applies to the `expectedPubKey.size()` comparison further down - the named constant makes the intent obvious.

src/sdk/main/src/ECDSAsecp256k1PrivateKey.cc
```diff
@@ -263,6 +305,53 @@ std::vector<std::byte> ECDSAsecp256k1PrivateKey::sign(const std::vector<std::byt
...
+  std::vector<unsigned char> compactSig(RAW_SIGNATURE_SIZE);
+  std::memcpy(compactSig.data(), r.data(), R_SIZE);
+  std::memcpy(compactSig.data() + R_SIZE, s.data(), S_SIZE);
+
+  // Get the expected uncompressed public key bytes (65 bytes: 0x04 || x || y).
+  const std::vector<std::byte> expectedPubKeyBytes =
+    ECDSAsecp256k1PublicKey::uncompressBytes(getPublicKey()->toBytesRaw());
+
+  // Create a secp256k1 context for verification/recovery.
+  secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
```
rwalworth: **suggestion (blocking):** This file already establishes a clean RAII pattern for crypto handles - every OpenSSL object that needs a custom deleter goes through [`OpenSSLObjectWrapper<T>`](src/sdk/main/include/impl/openssl_utils/OpenSSLObjectWrapper.h) and gets a thin derived class (see [`EVP_MD_CTX`](src/sdk/main/include/impl/openssl_utils/EVP_MD_CTX.h) for the closest analog: a context with no copier, just a custom free function). The `secp256k1_context` should follow the same pattern.

The current code does manual `secp256k1_context_create(...)` / `secp256k1_context_destroy(ctx)` pairing, which:
- Doesn't match the codebase's established RAII discipline for crypto handles
- Leaks the context if anything between create and destroy throws (defensive issue, not an active bug today)
- Doesn't give us a reusable type for any future secp256k1 calls we add - the next contributor will copy this manual pattern

Despite the name, `OpenSSLObjectWrapper` is purely templated - there is nothing OpenSSL-specific in its implementation. Two ways to apply it here:

**(a) Minimal churn:** add a new `src/sdk/main/include/impl/secp256k1_utils/` directory mirroring `openssl_utils/`, with a five-line `Secp256k1Context` that derives from `OpenSSLObjectWrapper<::secp256k1_context>` (move-only, deleter `&secp256k1_context_destroy`, no copier - same shape as `EVP_MD_CTX`). The slight awkwardness of inheriting from something called `OpenSSLObjectWrapper` is the only cost.

**(b) Cleaner:** rename the base to something neutral (e.g. `CHandleWrapper` or `CryptoObjectWrapper`), move it to `impl/`, and have both `openssl_utils/*` and the new `secp256k1_utils/*` derive from it. Each existing wrapper header becomes a 1-line edit (just changing the parent name); none of the actual semantics change. This is mechanical and leaves the codebase more consistent.

I'd take (b) - it's the right structural fit, and the rename is a sed-able change - but (a) is acceptable if you'd rather keep the diff small (a follow-up issue can be created to do (b)).

Either way, once `getRecoveryId` holds the context via the wrapper, the manual `secp256k1_context_destroy` goes away and the function gets shorter. If you also want to micro-optimize by reusing one context (the library docs recommend creating it once, and recovery+serialize only need `secp256k1_context_static` semantics), a function-local `static` instance of the new wrapper handles that too.

src/sdk/main/src/ECDSAsecp256k1PrivateKey.cc
```diff
@@ -263,6 +305,53 @@ std::vector<std::byte> ECDSAsecp256k1PrivateKey::sign(const std::vector<std::byt
...
+  if (!ctx)
+  {
+    return -1;
+  }
+
+  int result = -1;
+
+  // Try each recovery ID (0-3) and check which one recovers our public key.
+  for (int recId = 0; recId < 4; ++recId)
```
rwalworth, line 339: **thought:** Java's implementation comments that "on this curve, there are only two possible values for the recovery id" and iterates 0–1 only - recovery values 2 and 3 require `r ≥ n` (curve order), which is cryptographically negligible in practice.

You're being defensive iterating 0–3, which isn't wrong; just noting that in the unlikely worst case this does ~2x more recovery work than needed. Either approach is fine - just leaving this for awareness, no change requested.

rwalworth (changes requested):
Thanks for picking this up @aashu2006! I dug into the architectural question since this has been in the backlog for a while - looking at whether it's right to bring in a second crypto library just for recovery - and I'm convinced this is the correct path. Worth being honest about the framing first: the other SDKs (Java/BouncyCastle, Go/btcsuite, Swift/secp256k1.swift, JS/@noble, Rust/k256) all use a non-OpenSSL crypto library, but that's their *foundational* ECDSA stack - they didn't add one specifically for recovery, they were already there. C++ is the unique case: OpenSSL is enough for sign/verify/derive of secp256k1, and it's *only* recovery where the gap exists. So we genuinely are choosing whether to pull in a second library for one missing primitive.

Three options, in order of cost: (a) add a targeted library scoped only to recovery, (b) replace OpenSSL with bitcoin-core/secp256k1 across our entire ECDSA path, or (c) wait for OpenSSL. (c) isn't viable: OpenSSL 4.0 shipped April 14, 2026 without it, and the upstream tracking issue [openssl/openssl#18705](https://github.com/openssl/openssl/issues/18705) is on a 2-month "implement-or-close" timer (the only maintainer comment, from `t8m` in March, asks for community help and warns it'll be closed by ~2026-05-23 otherwise - no in-flight PR, no roadmap). Even an optimistic timeline puts availability multi-year out. (b) is a much larger refactor with real risk to working code paths. (a) - what this PR does - is the smallest-blast-radius path: bitcoin-core/secp256k1 is the reference implementation, well audited, narrowly scoped to a single primitive; the vcpkg port at `vcpkg/ports/secp256k1/` is already prepared with `-DENABLE_MODULE_RECOVERY` enabled; OpenSSL stays for everything else; and if upstream OpenSSL ever does add recovery, removing the dep is one file change.

The blockage is real and concrete: the TCK already calls `privateKey.getRecoveryId(r, s, message)` on its Ethereum tests, and `EthereumTransactionDataEip1559.mRecoveryId` exists as a field with no way to populate it. We can't get past those without this method.

Two blocking asks before merge:

1. **Add a known-vector test.** The current six tests only assert the result is in `[0, 3]` or `-1` - none check that for a *fixed* `(privateKey, message, r, s)` we get a *specific* expected ID. Without that, an endianness or off-by-one bug in the recovery loop would pass all six silently.
2. **Use the existing RAII pattern for the `secp256k1_context`.** The codebase has [`OpenSSLObjectWrapper<T>`](src/sdk/main/include/impl/openssl_utils/OpenSSLObjectWrapper.h) - purely templated, nothing OpenSSL-specific in its impl - and every crypto handle goes through it. The manual `secp256k1_context_create` / `secp256k1_context_destroy` pair in this PR doesn't match that discipline and leaks if anything between the two ever throws. A five-line `Secp256k1Context` derived from the existing wrapper (same shape as `EVP_MD_CTX`) is the right fit. See the inline comment for two ways to land this.

Smaller suggestions inline. Once the blockers are addressed I'll get this merged.

Also worth filing a follow-up issue: this PR adds the `getRecoveryId` primitive but doesn't yet wire it into `EthereumTransactionDataEip1559` / `EthereumFlow` to actually populate `mRecoveryId` during transaction construction. That's the work that closes the loop on issue #684's stated goal.

rwalworth (approved):
LGTM @aashu2006 - clean response to the review. Running the workflows now and will merge once they pass.

I'll file the follow-up issue for actually wiring `getRecoveryId` into `EthereumTransactionDataEip1559` / `EthereumFlow` - this PR adds the primitive, the next one closes the loop on #684's stated goal of "proper Ethereum transaction construction."

rwalworth (changes requested):
The clang-format-17 check is failing on `ECDSAsecp256k1PrivateKey.h` and `ECDSAsecp256k1PrivateKey.cc` - looks like the continuation indent on the multi-line `getRecoveryId` parameter list is off by a space, plus one of the `tryRecoverWithId(...)` call lines runs past the column limit. Both auto-fix with:

​```bash
clang-format-17 -i src/sdk/main/include/ECDSAsecp256k1PrivateKey.h src/sdk/main/src/ECDSAsecp256k1PrivateKey.cc
​```

Push that and we should be green.

rwalworth (approved):
LGTM, thanks @aashu2006!

PR #1633, opened 2026-05-10
feat: add Linux Clang+libc++ build configuration

### LINKED ISSUE #1620: [Intermediate]: Add `linux-x64-clang-libcxx` preset and run it as a nightly build
### 🧩 Intermediate Friendly

> [!IMPORTANT]
> ### 🧭 About Intermediate Issues
>
> Intermediate Issues are a great next step for contributors who enjoy digging into the codebase and reasoning about how things work.
>
> These issues often:
> - Involve multiple related files or components  
> - Encourage investigation and understanding of existing behavior  
> - Leave room for thoughtful implementation choices  
> - Stay focused on a clearly defined goal  
>
> Other kinds of contributions — from beginner-friendly tasks to large system-level changes — are just as valuable and use different labels.

### 👾 Description of the Task

The Linux Clang+libstdc++ preset (tracked in #1619) catches the most common Clang frontend regressions on PRs, but it does not exercise **libc++**, the standard library used on macOS. Apple Clang on `macos-latest` runners is the only nightly job that hits libc++ today, and a `macos-latest` build takes ~50 minutes.

Adding a Linux Clang+libc++ build gives us a much cheaper proxy for catching libc++-specific portability issues - `<filesystem>` differences, locale handling, allocator-aware container details, `std::format` / `std::ranges` quirks (where applicable to C++17), and template instantiation differences. It runs on the existing self-hosted Linux runners, so the runtime cost is bounded by Linux build time (~30 min) rather than GitHub-hosted macOS billing.

This is a nightly-tier job, not a PR gate, because:
1. The libstdc++ Clang job already catches Clang frontend issues at PR time.
2. libc++ ports often surface latent bugs that take iteration to fix; gating PRs on it would block unrelated work.
3. It requires a custom vcpkg triplet that rebuilds the C++ dependency tree against libc++ - first-time builds will be slow until the binary cache warms.

Relevant files:
```
CMakePresets.json
.github/workflows/zxc-build-library.yaml
.github/workflows/on-schedule-builds.yaml
vcpkg/triplets/  (or a project-local triplets directory)
src/sdk/main/**  (only if libc++ surfaces errors)
```

### 💡 Proposed Approach

Define a custom vcpkg triplet (e.g. `x64-linux-clang-libcxx`) that compiles all C++ dependencies with `clang++ -stdlib=libc++`. Without this, vcpkg dependencies built against libstdc++ cannot be linked into a libc++ SDK build (incompatible C++ ABI for libraries that expose C++ types across the boundary, such as gRPC, Protobuf, abseil, and log4cxx).

Add a new CMake preset that points at the custom triplet and applies `-stdlib=libc++` to the SDK build. Reuse all build flags and structure from the libstdc++ Clang preset; the only differences are the triplet name and the stdlib flag.

Wire the preset into the **nightly** build workflow (`on-schedule-builds.yaml`), not the PR workflow.

Fix whatever libc++-specific issues surface. Common categories:
- Header transitivity differences (libc++ tends to be stricter about what's transitively included)
- `std::filesystem` / `std::chrono` / locale behavior nuances
- `<algorithm>` / `<numeric>` template instantiation edge cases
- Differences in `noexcept` specifications between the two stdlibs

Avoid stdlib-conditional code unless absolutely unavoidable; prefer fixes that work cleanly under both libstdc++ and libc++.

Trade-offs:
- **Custom triplet vs `VCPKG_CHAINLOAD_TOOLCHAIN_FILE`:** custom triplet is cleaner and gives the binary cache a stable key. Chainloading is more flexible but harder to reason about.
- **Cold-build cost:** the first nightly run will rebuild the entire dep tree under libc++. Self-hosted runners with persistent vcpkg cache will warm up after one or two runs.

### 👩‍💻 Implementation Steps

- [ ] Create a custom vcpkg triplet `x64-linux-clang-libcxx`. Place it in a project-local triplets directory (e.g. `vcpkg-triplets/`) and set `VCPKG_OVERLAY_TRIPLETS` from the preset. The triplet should set `VCPKG_CXX_FLAGS=-stdlib=libc++`, `VCPKG_LINKER_FLAGS=-stdlib=libc++ -lc++abi`, and use `clang`/`clang++` as the compilers.
- [ ] Add new configure presets in `CMakePresets.json`:
  - `linux-x64-clang-libcxx-debug`
  - `linux-x64-clang-libcxx-release`
  Inherit from `vcpkg-base`, set `CMAKE_C_COMPILER=clang`, `CMAKE_CXX_COMPILER=clang++`, `VCPKG_TARGET_TRIPLET=x64-linux-clang-libcxx`, `VCPKG_OVERLAY_TRIPLETS=<path>`, and add `-stdlib=libc++` to `CMAKE_CXX_FLAGS_INIT`.
- [ ] Verify the preset configures and builds cleanly locally. Expect the first build to be slow while vcpkg rebuilds the dep tree under libc++:
  ```bash
  cmake --preset linux-x64-clang-libcxx-debug -DBUILD_TESTS=ON
  cmake --build -j 6 --preset linux-x64-clang-libcxx-debug
  ```
- [ ] Address any libc++-specific compilation errors. Keep fixes idiomatic across both stdlibs.
- [ ] Add a "Build (Linux, clang+libc++)" job to `.github/workflows/zxc-build-library.yaml`, gated by an input flag (e.g. `run-clang-libcxx-build`) defaulting to `false`.
- [ ] Update `.github/workflows/on-schedule-builds.yaml` to set `run-clang-libcxx-build: true` so the nightly run exercises this preset.
- [ ] Run the test suite under the new preset and confirm parity:
  ```bash
  ctest -j 6 -C Debug --test-dir build/linux-x64-clang-libcxx-debug -E NodeUpdateTransactionIntegrationTests --output-on-failure
  ```
- [ ] Update `CLAUDE.md` to document the new preset and triplet.

### ✔️ Acceptance Criteria

- [ ] `x64-linux-clang-libcxx` vcpkg triplet exists and successfully builds the C++ dependency tree
- [ ] `linux-x64-clang-libcxx-debug` and `linux-x64-clang-libcxx-release` presets exist and configure/build cleanly
- [ ] The new build job runs in the nightly workflow only — not on PRs
- [ ] All existing unit and integration tests pass under the libc++ build
- [ ] Any source-level fixes compile cleanly under libstdc++ and libc++
- [ ] Documentation (`CLAUDE.md`) lists the new preset and triplet
- [ ] No unrelated behavior or API changes are introduced

---

### 📋 Step-by-Step Contribution Guide

To help keep contributions consistent and easy to review, we recommend following these steps:

- [ ] Wait for assignment

Read [Workflow Guide](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/docs/training/workflow.md) for step-by-step workflow guidance.
Read [README.md](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/README.md) for setup instructions.

See the [Signing Guide](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/docs/training/signing.md).

### 🤔 Additional Information

Context and dependencies:
- **Depends on:** `add-preset-linux-clang-libstdcxx.md`. That issue establishes the Clang preset pattern in `CMakePresets.json` and the Clang CI job structure. This issue should land after it so the patterns can be reused.
- **Why nightly, not PR-blocking:** the libstdc++ Clang job already catches Clang frontend issues at PR time. The libc++ axis is about stdlib-portability bugs, which historically take iteration to fix and would unfairly block unrelated PRs.
- **libcxxabi:** the triplet may need `-lc++abi` explicitly depending on the runner's libc++ packaging. Verify against the `hiero-client-sdk-linux-large` runner image.
- **Binary cache warmup:** budget one or two nightly runs for the cache to populate before considering this build "fast." Once warm, expect runtimes comparable to the existing GCC nightly Linux job (~30 min).
- **Future consideration:** if libc++ surfaces no portability issues over several months, this is evidence we could de-emphasize the macOS nightly build (which is ~50 min on billed runners). That decision is out of scope for this issue.

If you have questions while working on this issue, feel free to ask! [Hiero-SDK-C++ Discord](https://discord.com/channels/905194001349627914/1337424839761465364)

hk2166:
**Description**:

Add Linux Clang+libc++ build configuration to test stdlib portability on self-hosted Linux runners, providing a faster and cheaper alternative to macOS builds for catching libc++-specific issues.

* Add custom vcpkg triplet `x64-linux-clang-libcxx` in `vcpkg-triplets/`
* Add CMake configure presets `linux-x64-clang-libcxx-debug` and `linux-x64-clang-libcxx-release`
* Add CMake build presets for both debug and release configurations
* Add `build-linux-clang-libcxx` job to `zxc-build-library.yaml` workflow
* Add `run-clang-libcxx-build` input flag to workflow (defaults to false)
* Enable libc++ build in nightly workflow via `on-schedule-builds.yaml`
* Document libc++ build requirements and usage in `README.md`

**Related issue(s)**:

Fixes #1620 

### Technical Implementation

**Custom vcpkg Triplet** (`vcpkg-triplets/x64-linux-clang-libcxx.cmake`):
- Configures vcpkg to build all C++ dependencies with Clang and libc++
- Sets compiler flags: `-stdlib=libc++`
- Sets linker flags: `-stdlib=libc++ -lc++abi`
- Ensures ABI compatibility across gRPC, Protobuf, abseil, log4cxx, and all other C++ dependencies

**CMake Presets**:
- Inherit from `vcpkg-base` for consistency
- Override compiler to Clang
- Add libc++ flags to compilation and linking
- Point to custom triplet via `VCPKG_OVERLAY_TRIPLETS`

**CI Integration**:
- New job runs only when `run-clang-libcxx-build: true` is set
- Enabled in nightly workflow, disabled for PR builds
- Installs `libc++-dev` and `libc++abi-dev` packages
- Builds and runs unit tests for both Debug and Release configurations

### DIFF
--- a/.github/workflows/on-schedule-builds.yaml
+++ b/.github/workflows/on-schedule-builds.yaml
@@ -14,4 +14,5 @@ jobs:
     with:
       run-windows-builds: true
       run-macos-builds: true
+      run-clang-libcxx-build: true
       upload-artifacts: false
--- a/.github/workflows/zxc-build-library.yaml
+++ b/.github/workflows/zxc-build-library.yaml
@@ -12,8 +12,14 @@ on:
         description: "Select this option when you want to run the MacOS builds."
         type: boolean
         required: false
-        default: false 
-       
+        default: false
+
+      run-clang-libcxx-build:
+        description: "Run the Linux Clang+libc++ build"
+        required: false
+        type: boolean
+        default: false
+
       upload-artifacts:
         description: "Select this option when you want to upload the artifacts."
         type: boolean
@@ -32,11 +38,17 @@ on:
         description: "Select this option when you want to run the MacOS builds."
         type: boolean
         required: false
-        default: false 
-         
+        default: false
+
+      run-clang-libcxx-build:
+        description: "Run the Linux Clang+libc++ build"
+        required: false
+        type: boolean
+        default: false
+
       upload-artifacts:
         description: "Select this option when you want to upload the artifacts."
-        type: boolean 
+        type: boolean
         required: false
         default: true
 
@@ -336,3 +348,61 @@ jobs:
           name: hapi-library-${{ runner.os }}-${{ steps.sha.outputs.short }}
           path: package/
           if-no-files-found: warn
+
+  build-linux-clang-libcxx:
+    name: "Build (Linux, clang+libc++)"
+    if: ${{ (inputs.run-clang-libcxx-build) && !cancelled() }}
+    runs-on: hiero-client-sdk-linux-large
+
+    env:
+      HIERO_NETWORK: localhost
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@a5ad31d6a139d249332a2605b85202e8c0b78450 # v2.19.1
+        with:
+          egress-policy: audit
+
+      - name: Checkout Code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          submodules: recursive
+
+      - name: Install libc++ and libc++abi
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libc++-dev libc++abi-dev
+
+      - name: Install CMake & Ninja
+        uses: step-security/get-cmake@082f0e11945209d5c8d8476a791e39d4d041005d # v4.3.1
+        with:
+          useCloudCache: true
+
+      - name: Setup VCPkg
+        uses: step-security/run-vcpkg@48490398616947b0df92dc1b24aab5339a66f9f0 # v11.5.2
+        with:
+          binaryCachePath: ${{ github.workspace }}/b/vcpkg_cache
+
+      - name: Convert VCPkg to Full Clone
+        working-directory: vcpkg
+        run: git fetch --unshallow --prune
+
+      - name: Configure CMake (Debug)
+        run: cmake --preset linux-x64-clang-libcxx-debug -DBUILD_TESTS=ON
+
+      - name: Build (Debug)
+        run: cmake --build --preset linux-x64-clang-libcxx-debug -j 6
+
+      - name: Run Unit Tests (Debug)
+        run: ctest -j 6 -C Debug --test-dir build/linux-x64-clang-libcxx-debug -R "TestVectors|UnitTests" --output-on-failure
+        continue-on-error: false
+
+      - name: Configure CMake (Release)
+        run: cmake --preset linux-x64-clang-libcxx-release -DBUILD_TESTS=ON
+
+      - name: Build (Release)
+        run: cmake --build --preset linux-x64-clang-libcxx-release -j 6
+
+      - name: Run Unit Tests (Release)
+        run: ctest -j 6 -C Release --test-dir build/linux-x64-clang-libcxx-release -R "TestVectors|UnitTests" --output-on-failure
+        continue-on-error: false
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -9,7 +9,7 @@
     {
       "name": "vcpkg-base",
       "hidden": true,
-      "displayName": "VCPkg Base Build",
+      "displayName": "VCPkg Base",
       "description": "Configured via the vcpkg toolchain",
       "binaryDir": "${sourceDir}/build/${presetName}",
       "installDir": "${sourceDir}/package",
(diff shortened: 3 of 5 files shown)


hk2166:
Hi @gsstoykov  @leninmehedy  could you please review my PR when you get a chance?

I’d also appreciate any feedback or suggestions for improvement. If there’s anything that should be changed or refined, I’m happy to work on it.

CMakePresets.json
```diff
@@ -9,11 +9,10 @@
     {
       "name": "vcpkg-base",
       "hidden": true,
-      "displayName": "VCPkg Base Build",
+      "displayName": "VCPkg Base",
       "description": "Configured via the vcpkg toolchain",
       "binaryDir": "${sourceDir}/build/${presetName}",
       "installDir": "${sourceDir}/package",
-      "generator": "Ninja Multi-Config",
```
rwalworth, line 16: **issue (blocking):** This change removes `"generator": "Ninja Multi-Config"` from the `vcpkg-base` preset. Since every existing Linux and macOS preset inherits from `vcpkg-base`, this drops them down to CMake's default generator on each platform (usually Unix Makefiles on Linux/macOS), which is a behavior change for the whole project - not just the new libc++ presets.

Beyond the runtime change, `gtest_discover_tests` is wired up with `WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}` (see [src/sdk/tests/unit/CMakeLists.txt:169](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/src/sdk/tests/unit/CMakeLists.txt#L169) and the integration equivalent). That `${CMAKE_BUILD_TYPE}` subdirectory is something multi-config generators produce; single-config generators don't, so removing this will likely break test discovery on the existing presets too.

Please restore the `"generator": "Ninja Multi-Config"` line in `vcpkg-base`. If you want the new libc++ presets to use a different generator, set it explicitly on those two presets only.

The display-name / cacheVariables shorthand cleanups in this file are fine - just the generator removal needs to be reverted.

README.md
```diff
@@ -107,6 +107,52 @@ cmake --preset linux-x64-release -DBUILD_TESTS=ON -DBUILD_TCK=ON -DBUILD_EXAMPLE
...
+
+- **First build will be slow** (30-60 minutes) as vcpkg rebuilds all C++ dependencies with libc++
+- Subsequent builds are much faster once the binary cache is populated
+- This configuration uses a custom vcpkg triplet (`x64-linux-clang-libcxx`) located in `vcpkg-triplets/`
+- All dependencies (gRPC, Protobuf, abseil, log4cxx, etc.) are rebuilt against libc++ to ensure ABI compatibility
+- This build runs in nightly CI to catch libc++-specific issues
+
+For more detailed information about the libc++ build configuration, see [CLAUDE.md](CLAUDE.md).
+
+For more detailed information about the libc++ build configuration, see [CLAUDE.md](CLAUDE.md).
```
rwalworth: **issue (blocking):** As we agreed on #1620, `CLAUDE.md` isn't a file we want in this repo - the mention in the issue's implementation steps was a copy-paste artifact from my template. README is the right home for all of this. Please remove both of these lines:

```md
For more detailed information about the libc++ build configuration, see [CLAUDE.md](CLAUDE.md).

For more detailed information about the libc++ build configuration, see [CLAUDE.md](CLAUDE.md).
```

(Also note they're duplicated - the same line appears twice in a row.)

Could you also update the PR description? It still says "Create `CLAUDE.md` with comprehensive developer documentation," which isn't accurate anymore.

.github/workflows/zxc-build-library.yaml
```diff
@@ -336,3 +350,61 @@ jobs:
...
+
+      - name: Configure CMake (Release)
+        run: cmake --preset linux-x64-clang-libcxx-release -DBUILD_TESTS=ON
+
+      - name: Build (Release)
+        run: cmake --build --preset linux-x64-clang-libcxx-release -j 6
+
+      - name: Run Unit Tests (Release)
+        run: ctest -j 6 -C Release --test-dir build/linux-x64-clang-libcxx-release --output-on-failure
+        continue-on-error: false
```
rwalworth, line 408: **issue (blocking):** The step is named "Run Unit Tests" but the `ctest` invocation here runs **every** test discovered for the preset - both unit and integration tests. Integration tests require a live Hiero network (see [config/local_node.json](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/config/local_node.json) and the `Prepare Hiero Solo` step in the existing `build` job above at [zxc-build-library.yaml:211](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/.github/workflows/zxc-build-library.yaml#L211)). Since this new job doesn't stand up Solo, the integration tests will fail on every nightly run as soon as the cache warms up.

The repo's [`README.md`](https://github.com/hiero-ledger/hiero-sdk-cpp/blob/main/README.md#L120-L124) already shows the right pattern for unit-only runs:

```sh
ctest -j 6 -C <CONFIG> --test-dir build/<PRESET> -R "TestVectors|UnitTests" --output-on-failure
```

You have two reasonable options:

1. **(Recommended for now)** Restrict the new job to unit tests by adding `-R "TestVectors|UnitTests"` to both `ctest` invocations. This matches the spirit of #1620 - the goal is libc++ stdlib portability for the SDK, and unit tests exercise that surface area without needing a network.
2. Wire in a `Prepare Hiero Solo` step before the test runs (matching the existing Linux job at lines 197-205) and use `-E NodeUpdateTransactionIntegrationTests` like the existing job does. This is a heavier nightly run but gives full coverage.

If you go with option 1, also rename the steps from "Run Unit Tests" to something accurate (the name is already correct in that case 🙂 - just keep it consistent).

src/sdk/tests/unit/TransactionUnitTests.cc
```diff
@@ -2773,7 +2773,7 @@ TEST_F(TransactionUnitTests, GetTransactionSizeWithSignatures)
                                                .setTransactionMemo("Test Memo")
                                                .setNodeAccountIds({ AccountId(3) })
                                                .setTransactionId(getTestTransactionIdMock())
-                                               .setKey(privateKey->getPublicKey())
+                                               .setKeyWithoutAlias(privateKey->getPublicKey())
```
rwalworth: **question (blocking):** What's the reason for this `setKey` --> `setKeyWithoutAlias` change? It looks unrelated to the build-configuration scope of this PR.

I checked, and:
- `AccountCreateTransaction::setKey()` is `[[deprecated]]`, but the project doesn't enable `-Werror=deprecated-declarations`, so the deprecation alone shouldn't break the new libc++ build.
- There are still ~40 other `.setKey(...)` call sites across the unit and integration test suites (e.g. `AccountCreateTransactionUnitTests.cc:131`, `AccountUpdateTransactionUnitTests.cc:154`, etc.). If this fix were genuinely needed for the libc++ build, you'd expect all of them to need it.

If something about the libc++ toolchain is actually erroring on this specific line, please share the error so I can understand the underlying issue (it would likely point at a real problem we'd want to fix more broadly). Otherwise, please revert this line - we can do a deprecated-`setKey` sweep as its own follow-up PR.

.github/workflows/zxc-build-library.yaml
```diff
@@ -12,13 +13,20 @@ on:
...
+        type: boolean
+        default: false
        
       upload-artifacts:
         description: "Select this option when you want to upload the artifacts."
         type: boolean
         required: false
-        default: true
+        default: true 
+
```
rwalworth: **nitpick:** Several of the new/changed lines in this file have trailing whitespace (e.g. `default: true ` on this line, `default: false ` on line 53, and a few of the blank "spacer" lines around the new `run-clang-libcxx-build` input). Also, there's a blank line just under `on:` near the top of the file that wasn't there before. None of this affects function - just an editor / pre-commit cleanup pass before next push.

.github/workflows/on-schedule-builds.yaml
```diff
@@ -14,4 +14,5 @@ jobs:
     with:
       run-windows-builds: true
       run-macos-builds: true
+      run-clang-libcxx-build: true
       upload-artifacts: false
```
rwalworth: **nitpick:** This file lost its trailing newline (`upload-artifacts: false` is now the last byte). Easy fix - add the newline back.

rwalworth (changes requested):
Thanks for picking this one up @hk2166! Solid skeleton - the custom triplet, presets, and nightly wiring all match the shape we discussed in #1620. A few things need to be fixed before this can merge - detailed comments below. Let me know if anything's unclear and I'll take another pass once these are addressed.

.github/workflows/on-schedule-builds.yaml
hk2166: Thanks for the detailed review. I’ve addressed the requested changes:

- Restored `Ninja Multi-Config` on the shared `vcpkg-base` preset.
- Removed the remaining `CLAUDE.md` references from the README and updated the PR description.
- Restricted the new libc++ CI job to unit/test-vector tests with `-R "TestVectors|UnitTests"`.
- Reverted the unrelated `setKeyWithoutAlias` test change back to `setKey`.
- Cleaned up trailing whitespace and restored the newline in `on-schedule-builds.yaml`.

hk2166:
Hi @rwalworth , thanks again for the detailed review — I’ve addressed the requested changes and pushed the updates.

Whenever you get a chance, could you please take another pass on the PR?

Also curious to hear your thoughts on the overall Clang + libc++ setup approach (custom triplet + presets + nightly-only validation) and whether the structure looks reasonable to you going forward.

